PR #615, opened 2013-12-03
Added possibility to create more types of derivative images, with any effects

Daniel-KM:
Hi,

I need to add another type of derivative image (a thumbnail of intermediate size with some effects).

Fortunately, all the code is ready for that in the core of Omeka: there is method to add derivative types to an uploaded file and a class Omeka_File_Derivative_Image_Creator.

Unfortunately, Omeka won't manage this new derivatives (creation and deletion) because they will not be a standard one. The problem is that the default derivatives are hard coded in application/models/File.php ($_pathsByType = [original, fullsize, thumbnails, square_thumbnails] and in the function createDerivatives() application/model/Files.

So I made a basic patch (see github) to make it more flexible. It:
- makes the PathsByType a function that get list of derivatives from database,
- adds an option in the appearance to list current derivatives and new ones, with their parameters,

The patch is not finalized (it should be integrated in the installer and in the bootstrap), but it is working. You just need to apply the patch and change options in Appearance > Parameters.

Do you think it's usefull or if there is a better way to do that?

I cross-post the patch on Github and Google, because I don't know where it is the better to talk about that.

Sincerely,
## 

Daniel Berthereau

### DIFF
--- a/application/controllers/AppearanceController.php
+++ b/application/controllers/AppearanceController.php
@@ -33,7 +33,6 @@ class AppearanceController extends Omeka_Controller_AbstractActionController
         $derivative_types = get_option('derivative_types');
         if (empty($derivative_types)) {
             $derivative_types = array(
-                'original',
                 'fullsize',
                 'thumbnail',
                 'square_thumbnail',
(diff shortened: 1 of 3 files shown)


cooperaj:
Just noticed that this was closed, but never merged. I'm looking at adding my own derivative types in 2.3.1 and this is far harder then it should be. I'm having to implement my own Storage adapter!

Am I doing it completely wrong?

Daniel-KM:
Hi,

I closed it, because I can't maintain a pull request with changes that occured after. Nevertheless, I use multiple derivatives with my omeka version (and some other improvements or change that are not yet accepted), that I try to keep as little as possible on https://github.com/daniel-km/omeka (about 20 commits). Anyway, this is just a way to add some micro, intermediate and big thumbnails, squared or not.

Sincerely,

Daniel Berthereau

PR #626, opened 2014-06-21
Added a call to parent search filters.

Daniel-KM:
Hi,

Some columns cannot be filtered, in particular the "id" column. So you can add either a case for it, either a call to parent method, as in this patch.

Sincerely,

Daniel Berthereau


Daniel-KM:
Hi,

The branch has been rebased on v2.5.

Sincerely,

Daniel Berthereau

PR #645, opened 2014-11-17
GD derivatives

Daniel-KM:
Hi,

I added a strategy to use GD, that is more often installed on shared hosts (more than external ImageMagick and Imagick extension). It is about 10-20% faster, but handles far fewer formats.

Sincerely,

Daniel Berthereau


zerocrates:
Why `imagecopyresized` for non-square thumbnails vs. `imagecopyresampled` for squares?

Daniel-KM:
Hi,

There is no difference, so I harmonized it.

Sincerely,

Daniel Berthereau

PR #647, opened 2014-11-26
Allowed image tag for any record.

Daniel-KM:
Hi,

Allowed file markup image tag for any record.

Sincerely,

Daniel Berthereau

### DIFF
--- a/application/views/helpers/FileMarkup.php
+++ b/application/views/helpers/FileMarkup.php
@@ -797,19 +797,14 @@ class Omeka_View_Helper_FileMarkup extends Zend_View_Helper_Abstract
         if (!$record) {
             return false;
         }
-        
-        if ($record instanceof File) {
-            $filename = $record->getDerivativeFilename();
-            $file = $record;
-        } else if ($record instanceof Item) {
-            $item = $record;
-            $file = get_db()->getTable('File')->getRandomFileWithImage($item->id);
-            if (!$file) {
-                return false;
-            }
-            $filename = $file->getDerivativeFilename();
-        } else {
-            // throw some exception?
+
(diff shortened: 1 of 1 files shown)


application/views/helpers/FileMarkup.php
```diff
@@ -797,19 +797,14 @@ public function image_tag($record, $props, $format)
...
-            $file = get_db()->getTable('File')->getRandomFileWithImage($item->id);
-            if (!$file) {
-                return false;
-            }
-            $filename = $file->getDerivativeFilename();
-        } else {
-            // throw some exception?
+
+        // Use the representative file.
+        if (!method_exists($record, 'getFile')) {
```
zerocrates: `getFile` is defined for all records, so I think it's better to do an argument typehint on `Omeka_Record_AbstractRecord` or a `instanceof` check for that class rather than duck-typing on the method name here.

Daniel-KM:
Hi,

Ok, I reset back to the original code, plus one check for <code>Omeka_Record_AbstractRecord</code> to allow any standard record.

Sincerely,

Daniel Berthereau

zerocrates:
Actually I liked your first change. You're right that all this logic is already handled now by getFile on the models, so we don't really need those specific checks there.

I just was suggesting that instead of your original `method_exists` check, we could just stick an `instanceof Omeka_Record_AbstractRecord` up on the initial check at line 797. The rest of the change was good (and would actually let us drop the explicit `->getFile()` call from `record_image`).

Daniel-KM:
Hi,

Ok, I didn't understand. I force pull an update of the commit.

Sincerely,

Daniel Berthereau

PR #651, opened 2015-01-28
Consistently show selected tags in search filter

annamichelle:
Changed tag_string to match tags selected from tag cloud. This ensures that tags selected from the Item Browse or Item Show pages are displayed in the Item Search Filter.

I had originally just added "tag" to ItemSearchFilter, but changing the global tag_string will be helpful in filtering exhibits, too.


zerocrates:
Can you rebase so we don't get the "first draft" commit?

annamichelle:
Should be fixed now!

PR #653, opened 2015-02-02
Fixed view script for plugins without view.

Daniel-KM:
Hi,

This is needed for some plugins like oai-pmh repository.

Sincerely,

Daniel Berthereau


zerocrates:
Since the variable never gets set at all in the case we're interested in here, can we switch that check from `!empty` to `isset`? Obviously `!empty` also works, but I'd like to be slightly more explicit.

Daniel-KM:
Hi,

Ok, I fix and rebased the commit.

Sincerely,

Daniel Berthereau

zerocrates:
Thanks!

PR #656, opened 2015-02-09
Accessibility

jerielizabeth:
Added skip nav and landmarks to admin views as well as to the default public views.


zerocrates:
A couple notes:

* There's no need to load jQuery manually in all those headers: it's already loaded by a call to `head_js`.
* A lot of these files (the CSS in particular) have a bunch of whitespace-only changes that make it hard to see the real ones. If your editor is set up to automatically strip or convert whitespace, please turn that feature off.
* We'll probably have to talk a little about some of the new text you introduced for aria-labels, whether those are the labels we want, and the translation of them.
* The JavaScript you're adding to the footer should be either contained in an existing external JavaScript file we're already using, or be in one of its own, so you don't have to write it in multiple places (the 2 in this PR is already one too many, and I assume from the structure that there are additional identical copies of that script in all the themes' footers in those PRs as well)

jerielizabeth:
@zerocrates @patrickmj I removed the javascript elements from the footers and the call in the headers. I also went through the themes and moved the javascript to the globals.js file, trying to follow existing patterns. 

What additional steps should I take to get the code ready for merge?

zerocrates:
So, I see a new commit that removes the inline scripts, but I don't see any equivalent to them for the admin theme, so it looks from here like that functionality has been removed from the admin. Was that intended?

jerielizabeth:
It is - the jquery was throwing errors with how I was loading it in the admin views (not in the themes though ... still figuring javascript out) so I removed it. The focus does update without the javascript in Chrome. I've been trying to test in Safari and Firefox, but for some reason can't seem to tab through the links in the most recent versions of those browsers for Mac. Still trouble shooting that, but it is functional without the javascript (at least for Chrome users...)

zerocrates:
I assume your issue had to do with your use of the dollar sign `$`. We tell jQuery to use "noConflict" mode, which doesn't set that `$` variable to anything. In the themes, you might be putting the code inside a function that sets the dollar sign up for you, but otherwise you need to refer to jQuery as `jQuery`.

It's curious that Chrome is working, because it's supposed to be one of the browsers that has trouble with skip links and correctly moving the keyboard focus so subsequent tabs work correctly.

jerielizabeth:
I was able to test with Firefox and Safari and it also worked in them without the javascript. I also was able to add the script back into the admin globals without error as a fallback for older browser versions and future browser troubles.

zerocrates:
* The application/views/scripts header is still manually including jQuery
* If that "skipnav" function is only intended to work on the skipnav link, I'd prefer to see that getting selected specifically: `$('#skipnav')` instead of all links with hrefs starting with `#`.
* I don't think the `+""` after `slice(1)` in that function actually does anything
* The common pattern seems to be to make the "Skip to content" link be visible when it's focused. It doesn't look like this does that
* Any user-visible (or audible) text needs to be getting passed through the `__` translation function: this includes link text and aria-labels. We need to upload those strings to translators, so we want to be sure these are the labels we're going with, and use common strings where possible ("Manage Site" vs. "Admin", do we want "Content" as the label for the main nav, etc.)

zerocrates:
Can we solve the `h1` problem by giving the `role=main` div an `aria-labelledby` pointing to the `h1`?

admin/themes/default/common/header.php
```diff
@@ -69,10 +69,10 @@
     <?php echo common('content-nav', array('title' => $title)); ?>
 
     <div class="subhead">
-        <?php echo search_form(array('show_advanced' => true)); ?>
+        <?php echo search_form(array('show_advanced' => true, 'form_attributes'=> array('role'=>'search'))); ?>
         <?php if (isset($title)) : ?>
-            <h1 class="section-title" title="<?php echo $title; ?>"><?php echo $title ?></h1>
+            <h1 id="content-heading" class="section-title" title="<?php echo $title; ?>"><?php echo $title ?></h1>
```
zerocrates: @kimisgold are you fine with this ID?

kimisgold: Works for me.

zerocrates:
@jerielizabeth is everything fine on your end after your changes Monday?

If so, I think this is pretty much good to go.

jerielizabeth:
@zerocrates Everything is the same on my end and reads well in jaws. The only remaining issue is the extra navigation element for the sidebar on mobile. I experimented with moving the ID up a level but it looked like it would need some additional styling changes. I can work on that as a separate pull request though.

PR #664, opened 2015-07-20
Fix spelling: "uncheking" should be "unchecking".

zploskey:
I noticed a typo in the English admin interface, which is corrected here. This effects most of the translation files, but does not change their meaning.

I went ahead and rebuilt the .pot and .mo files and included those. Please let me know if that's not the right approach. Perhaps someone with access to Transifex can push these changes up there.


zerocrates:
Thanks! The .mo/.po stuff has to be handled a little differently to work and not clobber peoples' updates on Transifex, so I handled that separately and just cherry-picked your first commit from here.

PR #665, opened 2015-07-21
Plural forms translation

jajm:
Hi,

a simple patch to allow plural forms translation.

Used like this:

```php
__(plural('one item', '%s items', $n), $n);
```

### DIFF
--- a/admin/themes/default/index.php
+++ b/admin/themes/default/index.php
@@ -2,16 +2,22 @@
 $pageTitle = __('Dashboard');
 echo head(array('bodyclass'=>'index primary-secondary', 'title'=>$pageTitle)); ?>
     
-<?php $stats = array(
-    array(link_to('items', null, total_records('Item')), __('items')),
-    array(link_to('collections', null, total_records('Collection')), __('collections')),
-    array(link_to('tags', null, total_records('Tag')), __('tags'))
+<?php
+$total_items = total_records('Item');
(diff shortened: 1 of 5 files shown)


zerocrates:
This looks pretty good to me.

We might need to do some investigation into how this will interact with Transifex and how it will interact with our existing translations before we can merge it, though.

zerocrates:
The only thing I can clearly see that I think is an issue with this is that the branch for "this is a plural string but translation is not enabled" should probably be using the passed-in number to pick between the singular and plural forms.

jajm:
I'm sorry I don't understand what you think is an issue here. What is that "branch" you are talking about ? It's Zend_Translate's job to pick between the singular and plural forms

zerocrates:
I'm talking about line 900 in your altered globals.php. That "else" branch only happens when there's no Translate object (which occurs when no custom locale is set). In that case, there's obviously no translator there to pick between the singular and plural source strings, so I believe the code there needs to do so itself (I assume this would look like `$string = ($msgid[2] === 1) ? $msgid[0] : $msgid[1];`).

A little tiny issue in the same place I just noticed as well: Our style calls for 4-space indentation, so lines 900 and 902 are under-indented.

jajm:
Oh I see. I will fix that immediately

zerocrates:
Okay, looks good.

PR #688, opened 2016-03-10
Addded 'starts with/ends with' advanced search options

luku:
I'm not sure if there are more people who need it, but achieving this simple feature via plugins/hooks/filters is really pain. 
I actually need only *starts with* option, but I'm guessing many will say, and where is *ends with* then? So I addded it for consistency sake as well.


zerocrates:
This seems like a fine addition to me.

luku:
As a promotional sidenote:) - *starts with* queries can use an index (range search)

zerocrates:
Just a note: this will probably take a little while to make it to a release, as it would be slated for 2.5.

PR #690, opened 2016-04-15
updated FileMarkup.defaultDisplay to use the linkText option

seweissman:
I was playing around with the FileMarkup helper, trying to get it to generate a text link, e.g.

    $html = $helper->defaultDisplay($file, array('linkText'=> 'Download', 'linkToFile' => true, 'linkToMetadata' => false));

I expected this to generate a link with "Download" as the text. Instead it generated a link with the original file name as the text. Looking at the code, the linkText option was not passed on from defaultDisplay to_linktoFile. I believe this change fixes this problem.


zerocrates:
Certainly looks like a bug

zerocrates:
Looks good; thanks!

Issue #689, opened 2016-04-07
Batch edit items - why not long running job?


luku:
I'm facing some difficulties with batch edit of more than a thousand items at once:

1. standard PHP settings for `max_input_vars` is 1000 - that's my responsibility to change
2. Omeka is running the batch edit job in default mode, so it's timing out - and here I'm not really sure what is best... 

I can see advantage of running the job in default mode, because I can throw and display exceptions to users and I don't need any page for showing the job status. But at the same time it's wrong to allow batch edit of few thousands items if the job cannot handle it. So I would like to get your feedback on this before creating some pull request. 
Do you think it's good idea to run the job in background? And if so, do we need some status page for job progress, or do you think silent processing with errors in log only is enough?

Also I think there are several ways for dispatching the job:
1. change the method for running job in `ItemsController` to `$dispatcher->sendLongRunning(`
2. add new option for it, i.e. `items_batch_edit_job_longrun` with `bool` value and decide based on that
3. decide how the job should run based on number of submitted items. For example `=< 1000 = default` and `> 1000 = long running`

Which one do you think is better, or other ideas??

As for status page, if it should be done, I have no idea where to put it... So any advice is appreciated. 
Thanks!

luku:
Thanks for the reply!
Yes, it's possible to achieve very high items/page by `per_page` query param, or in settings. 
I agree the majority of batch edits are about smaller number of items, but the changes needed are rather simple. That's why Bulk Metadata Editor plugin feels like an overkill, where we would need to hide or disable too powerfull/harmfull features from users:) But I also understand that running background script is not always possible, so maybe it's really something that plugin should do...
But please let me know if you come up with alternative that would fit Omeka core.

luku:
Adding real life scenario (for better motivation): Imagine you import few thousands items. After import you do some random changes. Now you visit your public page and see it's empty. Well, it's because all those thousands of objects were imported as private. What can I do? I did some changes in the imported items, so I don't want/cannot undo import. I just need to change them all to public. (Bulk Metadata Editor plugin cannot help with this). So you change your settings for items/page in admin to 900 (to make sure it's working) and change all private items to public in let's say 30 batch edit jobs. Then you visit your public page again. OH, NO! Why are not those items in my collection? Did I forget it during import??....
I think there's no need to continue, the frustration could affect good mood here:) It's easy to do this in DB directly, but it would be big help if users could do this themselves in admin UI.

zerocrates:
[...]

PR #694, opened 2016-05-05
Added a button to batch all browsed items.

Daniel-KM:
Hi,

This is a first commit to resolve the issue #689: a button is added to process all browsed items, filtered or not, via a background process. For example, this is possible to change the collection of a big number of items (see #693).

Two points may be improved: batch deletion and acl checks inside the background job.

Sincerely,

Daniel Berthereau

### DIFF
--- a/application/controllers/ItemsController.php
+++ b/application/controllers/ItemsController.php
@@ -286,6 +286,10 @@ class ItemsController extends Omeka_Controller_AbstractActionController
             throw new Omeka_Controller_Exception_403;
         }
 
+        if ($this->_getParam('all')) {
+            return $this->_batchEditAllSave();
+        }
+
         $itemIds = $this->_getParam('items');
         if ($itemIds) {
             $metadata = $this->_getParam('metadata');
@@ -380,14 +384,8 @@ class ItemsController extends Omeka_Controller_AbstractActionController
      *
      * @return void
      */
-    public function batchEditAllSaveAction()
+    protected function _batchEditAllSave()
     {
-        $hashParam = $this->_getParam('batch_edit_hash');
-        $hash = new Zend_Form_Element_Hash('batch_edit_hash');
-        if (!$hash->isValid($hashParam)) {
-            throw new Omeka_Controller_Exception_403;
-        }
-
         // Get the record ids filtered to Omeka_Db_Table::applySearchFilters().
         $params = json_decode($this->_getParam('params'), true) ?: array();
         $totalRecords = $this->_helper->db->count($params);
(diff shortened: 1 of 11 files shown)


application/controllers/ItemsController.php
```diff
@@ -248,7 +286,26 @@ public function batchEditSaveAction()
...
+        $editAll = (boolean) $this->_getParam('editAll');
+
+        if ($editAll) {
+            // Get the record ids filtered to Omeka_Db_Table::applySearchFilters().
+            $params = json_decode($this->_getParam('params'), true) ?: array();
+            $alias = $this->_helper->db->getTableAlias();
+            $select = $this->_helper->db
+                ->getSelectForFindBy($params)
+                ->reset(Zend_Db_Select::COLUMNS)
+                ->reset(Zend_Db_Select::LIMIT_COUNT)
```
zerocrates: Do you need to reset count and offset? `getSelectForFindBy` shouldn't be applying the pagination in the first place, it's just `findBy` that does that.

Daniel-KM: You 're right.

zerocrates:
Interesting! There are definitely quite a few people who'd like this option.

My major reservation is that I think this would work better as a separate job (and separate save dialog), rather than behavior in the existing ones that's switched on by `$editAll`. Having a little more separation would make it so you could just pass the params straight to the "edit all" job, which would do the "find by" query in the background and not have to deal with a big list of IDs at all in the way the current one does. Apart from that, I think separation of the two types of batch edits would lead to cleaner code in the controller and the job(s).

I'll have to look more carefully at how this works, but it does look promising.

Daniel-KM:
Hi,

I replaced the button by a checkbox and created a new job for edition and deletion.

Sincerely,

Daniel Berthereau

zerocrates:
I've had a chance to take a look at actually using the branch and it looks good! I have a few things I noticed, though:

The ACL loading that you're doing, it might be a good idea to do that _before_ you run the query to get the items. If the ACL is there when the findBy gets run, we'll add the correct params to exclude items the current user can't see. With the permssions we actually have this is probably more of an academic distinction as people who can edit tend to be able to see everything, but it's a good idea nonetheless.

Lower-level users currently get a 403 when trying to do an "all" batch edit. It looks like they don't have permissions for the batch-edit-all-save action.

The redirect that happens after doing an "all" batch transforms the params into the Zend slash-separated style (`/browse/sort_field/added/sort_dir/d/search/`, etc.) and at least in some cases this messes up the ItemSearchFilters helper. Can you make the redirect instead use the query string?

Daniel-KM:
Hi,

I reset the rediction and fixed the acl for non admin user.

Sincerely,

Daniel Berthereau

Daniel-KM:
Hi,

The new UI is better and fine for me. 

I improved the logs in the last commit (there is only one modified file and there is no conflict despite the message just above).

Sincerely,

Daniel Berthereau

Daniel-KM:
See #729.

PR #695, opened 2016-05-09
Addded regex in advanced search options.

Daniel-KM:
Hi,

I added advanced search options in the admin view to search items. 

In fact, I'm currently finishing to improve Bulk Metadata Editor (https://github.com/Daniel-KM/BulkMetadataEditor/tree/background_job): I already added a job process, but the ajax display is still slow for big queries and regex are needed to avoid to duplicate code. 

Sincerely,

Daniel Berthereau


zerocrates:
We might need to change the user-facing labels for these to something that's a little more obvious to people who won't immediately recognize what "regexp" means. Something with "matches" would help, I think, like "matches expression" or "matches regular expression" or even possibly just "matches."

zerocrates:
Otherwise I don't see any particular reason not to include regex matches among the advanced options.

Daniel-KM:
Hi,

I renamed options. Note that I don't add them by default in the public view.

Sincerely,

Daniel Berthereau

zerocrates:
As to that last commit, on a blank regex search it's probably best to just do nothing (just `continue 2;`). If the user wants to search on emptiness/NULL there's already a query type for that.

I think "matches" works fine for the label. On the other hand, I don't think it's a good idea to change "is not exactly" to "is not". I'd prefer to just leave that as it is (and at any rate it's not related to the rest of this change).

Daniel-KM:
Hi,

I reverted to use continue 2.

Sincerely,

Daniel Berthereau

PR #696, opened 2016-05-10
Correct checkbox id for unselected record types in search menu

paxed:
Uncheck any of the record types in the header search menu, and
do a search. The record types which were unselected are now generated
with id="record_types" instead of id="record_types-File" for example,
and the corresponding label does nothing.

### DIFF
--- a/application/views/scripts/items/search-form.php
+++ b/application/views/scripts/items/search-form.php
@@ -103,7 +103,7 @@ $formAttributes['method'] = 'GET';
         </div>
     </div>
 
-    <div class="field">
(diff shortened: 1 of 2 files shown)


zerocrates:
I see what's going on with the record type checkboxes and that change looks good, but what's up with the second, seemingly unrelated commit on here?

paxed:
Whoops, my bad - I messed up the commit, the other one should've been for pull req #697 ... I'll fix it tomorrow.

paxed:
Wrong branch, correct is in pull req #698

PR #697, opened 2016-05-10
Add the missing id attributes to item search form

paxed:


zerocrates:
@kimisgold any objection to this?

PR #698, opened 2016-05-11
Correct checkbox id for unselected record types in search menu

paxed:
Uncheck any of the record types in the header search menu, and
do a search. The record types which were unselected are now generated
with id="record_types" instead of id="record_types-File" for example,
and the corresponding label does nothing.


zerocrates:
Thanks!

PR #702, opened 2016-05-20
Move HTML for items search filters into partial view, to allow theming.

luku:
Possible to modify the view both in public and admin theme

### DIFF
--- a/application/views/helpers/ItemSearchFilters.php
+++ b/application/views/helpers/ItemSearchFilters.php
@@ -104,22 +104,9 @@ class Omeka_View_Helper_ItemSearchFilters extends Zend_View_Helper_Abstract
             }
         }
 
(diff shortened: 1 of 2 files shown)


zerocrates:
What's the reason behind passing the requestArray along?

luku:
Actually I think I will close this pull request... My ultimate goal was to modify the search filters in the way that will allow you to remove single filter as well (via link). That's why I passed the requestArray along, and hoped it will allow me to do it. But later I realized, it doesn't work, firstly due to the filter for displayArray, secondly because the current key-value pairs, doesn't support simple expansion, needed for this. 
So now I'm considering writing whole new helper  instead. 
If you think this pr has some value, I can remove the requestArray to strictly follow the original code

PR #704, opened 2016-05-24
Use HTML5 datalist to autosuggest searchable tags

paxed:
The tag search input in advanced search form is pretty useless
as-is; users have no easy way to know what tags are available
(without looking in the tag cloud).  Use the HTML5 datalist
element to allow the input autosuggest a tag.

This is perhaps not the best way to generate the datalist...


zerocrates:
I think we'll have problems with `datalist` not being supported by browsers Omeka supports (like Safari and older versions of IE).

The "Add Tags" input on item edit uses an already-fleshed-out jQuery-based autosuggester that goes to the server using AJAX for suggestions. Maybe you could use that instead?

PR #710, opened 2016-05-31
Closing tag must be same as opening

luku:


zerocrates:
Obviously this wasn't a mistake; `<video> </audio>` is just experimental next-gen HTML that the world isn't ready for yet.

PR #713, opened 2016-06-04
removed unnecessary else statements

royopa:


zerocrates:
Thanks for the pull request.

I don't think I'll merge this one, though. There's nothing wrong with it per se, but particularly in "plumbing" code like the bootstrap file, I'd prefer not to make changes like this that are basically cosmetic. There's definitely a line somewhere where reducing indentation and clauses is a worthwhile change by itself, but I think this particular change falls on the wrong side of that line.

zerocrates:
While we didn't make this particular change, we have simplified the bootstrap somewhat by other means, removing some variables and conditionals and consolidating others.

As my personal preference is still to maintain the `else` clauses and our other changes have made it so this PR no longer cleanly applies, I'm going to officially close this request.

PR #715, opened 2016-06-09
Toggle item details should work on first click

luku:


zerocrates:
Yep, that's clearly backwards in the current code. Surprised it wasn't noticed long ago, but I suppose people just thought they mis-clicked the first time (and/or don't use that button often, which is fairly likely).

PR #718, opened 2016-07-01
Allow zooming in admin for phones/tablets.

luku:
Not sure why was the maximum-scale there, but I couldn't see any weird behavior after removing it.


zerocrates:
@kimisgold any issue here?

kimisgold:
Nope.

PR #722, opened 2016-07-18
Added a warning when leaving a modified form without save.

Daniel-KM:
Hi,

I added a warning when leaving a modified form without save when adding or editing items, collection and files.

Sincerely,

Daniel Berthereau


zerocrates:
We have an implementation of this in S. I'll have to take a longer look to see if it's better/simpler for us to backport that.

Otherwise my one concern is with doing this on just a subset of forms like this. On the one hand it will help people out with a common problem, but on the other I'm worried that having it in _some_ places sets up the expectation for users that it will be there on other complex forms (like the Exhibit Builder's, for example) and it might worsen our existing problems in those places.

Daniel-KM:
Hi,

I'm going to add it in all forms of Omeka, because it can be added anywhere easily (just a javascript include and call). But I can't add it automatically for all plugins, because sometime, there is an ajax modification of an input field, and in that case a trigger should be added internally. 

Sincerely,

Daniel Berthereau

Daniel-KM:
Hi,

I updated the code so now any form can be "are-you-sure", just by adding the class "warn-no-save". 

In fact, there were two possibilities for the default: all the forms except one class (for example for the header search form), or no form except the specified ones. I prefer the latter, because there can be a line or two to add for ajax forms. Anyway, this is just a class to set in globals.js, so you can change it easily.

I can add it to the two main plugins Simple Pages and Exhibit Builder, but this is useless if this is not included in the core.

Sincerely,

Daniel Berthereau

Daniel-KM:
Hi,

I rebased the branch on the last master to fix a merge issue.

Sincerely,

Daniel Berthereau

CathyHajo:
This is a fix that we badly need in all versions, not just in Omeka-S. We've lost complex records by accidentally navigating away when we meant to open a new tab. It's odd that Omeka warns you when you remove an added field, but not when you are about to leave the whole item record without saving.

zerocrates:
I'm not so sure about the way this library works where it uses the "change" events for the form elements when its deciding if the form is "dirty." We've had pretty good luck with basically the opposite strategy on S, just using a `serialize` call at beforeunload.

You can see where the "are-you-sure" way requires a little more custom code in the areas where you're having to manually tell it to "check" the form. Neither strategy will work with no form-specific tweaks whatsoever, but I think it'd be best to keep them to a minimum.

Daniel-KM:
Hi,

The "are-you-sure" library is simple too: just add a class and the form is ready to warn. A callback may be added for Ajax elements. It manages not only changes made by Ajax, but changes that doesn't matter too (add a class on these elements). The serialize method doesn't manage this case directly and there is no perfect solution.

Anyway, the question is: Do you have the time to implement the Omeka S solution or is it better to merge now and to revert this branch later when you or somebody else will have some time? 

Sincerely,

Daniel Berthereau

luku:
I did similar warning, that is based on jQuery serialize as well, with additional care for `input[type=file]`. Snippet:
``` js
		var states = [$form.serialize()];
		// $.serialize() doesn't include <input type=file tags at all
		$form.find('input[type=file]').each(function() {
			var el = $(this);
			states.push(el.attr('name') +'='+ el.prop('value'));
		});
```
Then `onbeforeunload` I compare the states when the form was loaded with current states. It seems to work pretty well, didn't receive any bug reports so far. 
The biggest weakness of this whole concept is that initial form state must be collected after all elements are loaded/inserted into DOM. Waiting for `omeka:elementformload` event was not enough, and I had to add 1000ms timeout before collecting it. This may break for other Omeka users, if they are filling their forms via AJAX or something else, when this 1sec timeout is not enough. 
Although this pull request is using jquery plugin, I'm worried it will not work as expected for us, since it's calling `Omeka.areYouSureCheck();` directly. 

Since this feature is in very sensitive places, isn't it better as plugin then? I mean if it's core feature we cannot simply disable it, or choose in which forms it will be used, without changing core views...

Daniel-KM:
Hi,

@luku 
The `Omeka.areYouSureCheck()` is called automatically, but it does nothing when there is no form with the class "warn-no-save". 

With the implementation of the jQuery plugin "are-you-sure" I did, only a class (and a callback when there are Ajax elements) should be added to warn. There is no need to wait a random time and it works with files too.

I can add a config option to make it enabled or disabled by default as a whole (simply with a main body class "warn" or "no-warn").

In fact, the main issue is that Omeka Classic doesn't allow a specific admin theme currently, although the code is nearly ready for that. So it's very difficult to do a plugin to manage this point. I like to change some other points, in the items/browse view too (new columns, or replacement of     "Details Edit Delete" by hover icons), but it's not possible currently.

Anyway, the feature is important, and the choice between "are-you-sure" or "serialize" is minor for me.

Sincerely,

Daniel Berthereau

luku:
Hi,
@Daniel-KM

> With the implementation of the jQuery plugin "are-you-sure" I did, only a class (and a callback when there are Ajax elements) should be added to warn. There is no need to wait a random time and it works with files too.

Ok, but from what I can see, it will require calling `Omeka.areYouSureRescan()` in every plugin that is loading something into form. It will not work as is, I guess. Safer than random timeout, I agree, but will require changes in related plugins.

> I can add a config option to make it enabled or disabled by default as a whole (simply with a main body class "warn" or "no-warn").

Yes, this sounds great!

Best regards,
Luk

Daniel-KM:
Hi,

@luku 
Do you have published your implementation of serialized? If so, @zerocrates may use it if he prefers.

Sincerely,

Daniel Berthereau

luku:
The snippet I gave is actually the whole thing, but I created [gist](https://gist.github.com/luku/56b2c7371708c5534ce79683f89232f6) with the whole logic for reference.
It doesn't work with multiple forms on single page, and it doesn't let you rescan form (because it relies on the timeout). I could rewrite it as jquery plugin, but not sure when...

Daniel-KM:
Hi,

I added an option in appearance/settings to enable/disable the warning, and updated globals.js functions so nothing is done when the option is disabled, even when there is a rescan().

Sincerely,

Daniel Berthereau

Daniel-KM:
Hi,

I have rebased the branch on last upstream/master to take care of a new merge issue.

Sincerely,

Daniel Berthereau

zerocrates:
Thanks for the update Daniel.

Apologies for the delay; we haven't forgotten about this.

zerocrates:
I've gone with the S implementation I'd mentioned before, so I'll be closing this.

Regardless, thanks for the work you put into this implementation and your advocacy for the feature in general.

PR #723, opened 2016-07-20
Allow "transparent" as a valid value.

seantwalsh3:


zerocrates:
The "transparent" option is missing the start-of-string and end-of-string anchors, so it will validate anything with "transparent" in it anywhere, opening a big hole in the validation.

zerocrates:
Also, the validation message has to be updated to correctly reflect what values are legal.

zerocrates:
Just out of curiosity, what's your use-case for having `transparent` in the custom color selector?

seantwalsh3:
I'm creating a custom theme for a client who wish to make it public once it is finalized. One of the key things this theme will do is allow for a lot of customization via the theme configuration. In some cases I can imagine that having transparent colors would be needed to implement designs with background graphics.

While a user can leave a color field empty it ultimately defaults to what the theme provides vs. being transparent.

zerocrates:
@kimisgold are we doing any programmatic lightening/darkening of user-set colors that would have to take account of the ability to set "transparent"? I seem to remember that we are somewhere.

kimisgold:
<a href="https://github.com/omeka/theme-thanksroy/blob/master/functions.php#L11-L27">We do in the Thanks, Roy theme.</a> It'd be pretty simple to accommodate, especially since it only occurs in one theme.

zerocrates:
Fixing Thanks, Roy to understand "transparent" won't be difficult, but I'm concerned about the many themes out there that are local edits of Thanks, Roy or otherwise are based off of it.

If allowing "transparent" was triggered by an option on the validator (with the default being to not allow it), that would avoid any possible problem as any old theme that couldn't deal with it would also not specify the option.

seantwalsh3:
Completely understandable, I can work on making it triggered via an option on the validator.

zerocrates:
Thanks! Feel free to ask for advice if you think you need it. As I remember, validator options are pretty straightforward but there's just some boilerplate involved (accepting both arrays and Zend_Config objects at least).

zerocrates:
I'm trying to get things wrapped up for 2.5, so if there's no movement here by the end of the week, I'm going to push this off of the 2.5 release.

PR #727, opened 2016-07-28
Disabled HTML filtering in TinyMCE editor

doana:
Hello, 

It seems problematic to me to have HTML filtering happening at two places in the codebase (TinyMCE + HTML Purifier). 

I believe it creates a confusing experience for the user because there is an option in the security settings to enable / disable HTML filtering, but it seems like it’s not respected because TinyMCE is doing its own filtering. 

Besides, TinyMCE isn’t an awesome solution on its own for HTML filtering, since it’s client side and can be easily bypassed.

I think the best solution would be to rely on HTML Purifier alone. 

What do you think? 

-Adam

### DIFF
--- a/admin/themes/default/javascripts/globals.js
+++ b/admin/themes/default/javascripts/globals.js
@@ -27,7 +27,8 @@ if (!Omeka) {
             media_strict: false,
             width: "100%",
             autoresize_max_height: 500,
-            entities: "160,nbsp,173,shy,8194,ensp,8195,emsp,8201,thinsp,8204,zwnj,8205,zwj,8206,lrm,8207,rlm"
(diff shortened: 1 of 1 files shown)


zerocrates:
I'm not opposed to it in principle. From a security standpoint you're right, we're not relying on TinyMCE's editing at all, because it's so trivially bypassable, and that's why we have HTML Purifier there.

The only thing I'm not sure of is whether verify_html does anything convenient for users that we'd lose by deactivating it... but I'm not seeing much in the way of evidence that that's the case.

If we did this, I believe there's some other settings/plugins we could stop using because they're just there to get the filtering to behave a little better anyway (for example to make it allow certain kinds of embedded media).

zerocrates:
This also reminds me that I need to look again into what's involved in moving us off the dead 3.x branch.

Obviously that's a different issue than this.

PR #733, opened 2016-08-18
Added an option to select the admin theme in Appearance/Settings.

Daniel-KM:
Hi,

I added an option to select the admin theme in Appearance/Settings (the option is displayed only when there are multiple themes).

This quick patch allows to customize some pieces of the admin theme simply by a copy of it in another folder in admin/themes. 

There are some works to do to make the class Theme more generic and to use the files in the theme default as default, but this is not really needed.

Sincerely,

Daniel Berthereau

### DIFF
--- a/application/forms/AppearanceSettings.php
+++ b/application/forms/AppearanceSettings.php
@@ -63,6 +63,17 @@ class Omeka_Form_AppearanceSettings extends Omeka_Form
             'class' => 'checkbox',
         ));
 
+        $adminThemes = Theme::getAllAdminThemes();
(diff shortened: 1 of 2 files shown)


zerocrates:
So this doesn't have any impact at all outside that settings page, the only place where we'll enumerate that directory, and the vast majority of users who haven't made or installed a different admin theme won't see any change at all.

If I'm thinking correctly then this looks good to me. Not anything I'd call a priority but it's basically costless for people not using it and pretty useful for people who would.

PR #734, opened 2016-08-23
Fixed display of themes/browse.

Daniel-KM:
Hi,

Fixed display of themes/browse.

Sincerely,

Daniel Berthereau


zerocrates:
What's the problem that this fixes?

Daniel-KM:
Hi,

In line 48, $current is used instead of $theme. Line 42, $current is used, but there is a condition line 33 so the same condition in line 42 is useless, so I rewrite the first condition line 33. I added an id to the form too.

Sincerely,

Daniel Berthereau

PR #750, opened 2016-10-20
Missing ACL check for deleting an item on edit page

luku:


zerocrates:
I believe this was omitted because with the normal permissions, deletion and editing of items go hand in hand (since after all, someone who can edit the item can blank out its contents).

Still, I don't see any harm in making this change for increased accuracy in the UI for the case where someone has altered the default permissions. I assume this same problem goes beyond just the Items though, at least to Files and Collections as well?

luku:
Yes exactly. We discovered missing permission check after we created new plugin where you can define custom permissions for contributors. 
Files already have the permission check there, so I'm adding it to Collections as well.

zerocrates:
OK, good, we're on the same wavelength. It makes sense to do these checks.

Thanks!

(My apologies for delayed responses to several of your PRs, it's been a busy time in the past month)

PR #753, opened 2016-10-24
Are you sure? Warning when leaving page

luku:
I merged Daniel-KM pull request #722 and simplified drastically the jQuery plugin as I promised.
It doesn't fire when the form gets dirty, this is simply checking when leaving page if anything in the form has changed. 
It works almost like the original plugin. Every plugin needs to call `Omeka.areYouSureRescan();` in the case it is adding some new form fields/values after the page has loaded (or after `omeka:elementformload` event is triggered). `Omeka.areYouSureRescan();` is simply updating the original form state that will be compared against, when leaving page. 
`Omeka.areYouSureCheck();` is removed.
It is possible to ignore warning, by default when deleting. 
I also added method/event for destroying the feature on fly (via `$('form').trigger('destroy.areYouSure');`)

So if Omeka team doesn't think real-time feedback about the changes in form are necessary, this is an alternative.


luku:
The original jquery plugin supports excluding certain fields from serialization (using i.e. class `ays-ignore`, `aysIgnore` or data attribute `data-ays-ignore`) on fields, that may change but we don't care about those changes). So I added support for this as well.

zerocrates:
As I said about the other PR on this, since the S variant is merged, I'll close this one. Thanks for your attention to that branch as well as your work on this one.

PR #757, opened 2016-11-10
Improved the detection of common mime types for some plain text.

Daniel-KM:
Hi,

The "text/plain" mimetype is not very useful, so this patch sets the common text/plain, in particular for json and xml files.

Sincerely,

Daniel Berthereau

### DIFF
--- a/application/libraries/Omeka/File/MimeType/Detect.php
+++ b/application/libraries/Omeka/File/MimeType/Detect.php
@@ -128,6 +128,28 @@ class Omeka_File_MimeType_Detect
             $this->_mimeType = (false === $fallbackMimeType) ? 'application/octet-stream' : $fallbackMimeType;
         }
         
+        // Detect some common "plain text" via the extension. Most important
+        // ones are json and xml.
+        if ($this->_mimeType == 'text/plain') {
+            $extensions = array(
+                'css' => 'text/css',
+                'csv' => 'text/csv',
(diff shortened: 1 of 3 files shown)


zerocrates:
This is only quasi-related to your change, but I noticed you removed the "handling" of mime_content_type being missing that mentions deprecation. The PHP manual _used_ to say mime_content_type was deprecated, but no longer does, so that makes sense.

The weird part is this: mime_content_type is provided by... ext/fileinfo! It doesn't seem like it makes sense to have mime_content_type and Fileinfo as separate "strategies" when their presence and results will always go together. There _is_ the minor possibility of Fileinfo being disabled or missing on systems where even the default extensions get pulled out into separate packages, so that might be an issue with removing the check for mime_content_type entirely.

zerocrates:
On a more direct note, what's the reasoning for having this check apply here in this particular strategy and not in Detect itself so it applies no matter which strategy is used?

Daniel-KM:
It works the same, because it's not possible to modify the list of strategies. But, this is more logical to move the process inside Omeka_File_MimeType_Detect / line 122. Do you want I commit it?

zerocrates:
I merged the mime detection part of this (the first two commits).

The class change isn't strictly related and it seems like there's a little room for improvement on it anyway (to do a single `str_replace` at least). It's still welcome as a change, though.

PR #760, opened 2016-11-25
Filter multiple item types and collections

Daniel-KM:
Hi,

I need to filter multiple item types and multiple collections, so I updated the filter, and allowed to check if no type is set, like for collection.

Sincerely,

Daniel Berthereau

### DIFF
--- a/admin/themes/default/collections/browse.php
+++ b/admin/themes/default/collections/browse.php
@@ -13,11 +13,12 @@ echo flash();
         </a>
     <?php endif; ?>
     <p class="not-in-collections">
-    <?php if ($totalItemsWithoutCollection): ?>
-        <?php $withoutCollectionMessage = __(plural('%sOne item%s has no collection.', "%s%d items%s aren't in a collection.", $totalItemsWithoutCollection), '<a href="' . html_escape(url('items/browse?collection=0')) . '">', $totalItemsWithoutCollection, '</a>') ?>
-    <?php else: ?>
-        <?php $withoutCollectionMessage = __('All items are in a collection.'); ?>
-    <?php endif;?>
+    <?php if ($totalItemsWithoutCollection):
+        $withoutCollectionMessage = __(plural('%s%d item%s has no collection.', "%s%d items%s aren't in a collection.",
+            $totalItemsWithoutCollection), '<a href="' . html_escape(url('items/browse?collection=0')) . '">', $totalItemsWithoutCollection, '</a>');
+    else:
+        $withoutCollectionMessage = __('All items are in a collection.');
+    endif; ?>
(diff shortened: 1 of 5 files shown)


zerocrates:
Expanding the multiple and null support sounds like a good idea to me. I have a couple concerns about consistency though:

The collection and item type filters seem to use opposite strategies for arrays: the collection version seems more appropriate as it checks each value, where the item type version is just checking the type of the first value. The two are doing the same thing, so they should do so the same way.

Similarly the check for "null" isn't quite the same between the two: the item type filter includes the empty string as a "null" choice while the collection one doesn't. It's probably best to not treat empty string as null, but it might make sense to treat actuall `null` as null.

Finally, this change is causing a test failure: `Models_Table_ItemTest::testSearchFilters`. I assume it's complaining because of the new cast to integer in this code, so just changing the test to not expect quotation marks around the value in the SQL should be a sufficient fix.

Daniel-KM:
Hi,

I committed a fix to improve consistency and added some tests. The `null` is not treated as an empty value, because it's not consistant with Zend.

Sincerely,

Daniel Berthereau

zerocrates:
One thing I notice doesn't work here is combining the "null" support mapped to zero with the "array" support: so you can't do a single search to get items in either Collection X or no collection.

I think the array versions added here should support all the same options as the one-at-a-time filter, both to avoid confusion and also to make it easier to eventually implement an interface to this support by just duplicating the existing `select` controls from the advanced search form.

Unfortunately IS NULL can't happen within an IN, so it'd need a little logic to see a zero in the array, drop it, and add a `collection_id IS NULL` (or `OR collection_id IS NULL` as appropriate).

It may end up simplest to do Collection and Item Type to _only_ have the code for dealing with an array (just wrapping an array around any non-array value given), since it has to duplicate all the simple-scalar logic anyway.

Daniel-KM:
Hi,

The simplest solution is to remove the NULL from the items, collections and files tables and to use only "0", but it may have some side effects. Is there any reason to keep Null?

Sincerely,

Daniel Berthereau

Daniel-KM:
Hi,

I genericized the filters and allowed all possibilities and added some tests.

Sincerely,

Daniel Berthereau

zerocrates:
The "null in array" check for item types won't work because the join there is an inner join.

Daniel-KM:
Hi,

I forgot to push the last fix.

Sincerely,

Daniel Berthereau

zerocrates:
In the interest of actually getting the 2.5 release out the door I didn't pull this in for that branch, but we should be able to merge it into master soon after that release (I've had parallel changes fixing the filter display side floating around for quite a while as well).

Issue #724, opened 2016-07-20
Allow users to select the thumbnail type used in the theme


zerocrates:
This is really more of a metaissue spanning the themes, but this is the closest thing we've got to a place to put it.

Switching from most themes' default choice of the square thumbnail to the "regular" preserved-aspect thumbnail is one of the most common things we see people doing or asking for support to do. It's a simple change at the theme code level, but that's an unnecessary burden for what should be a pretty trivial change.

Adding a switch to theme configuration for this should be very simple. The only likely problematic piece will be theme styles that assume "squareness" of the images.

PR #763, opened 2016-12-05
Implement “use square thumbnail” appearance option and apply to defau…

kimisgold:
…lt views that use thumbnails. (#724)

### DIFF
--- a/application/forms/AppearanceSettings.php
+++ b/application/forms/AppearanceSettings.php
@@ -20,6 +20,7 @@ class Omeka_Form_AppearanceSettings extends Omeka_Form
 
         $this->addElement('checkbox', 'use_square_thumbnail', array(
             'label' => __('Use Square Thumbnails'),
+            'description' => __('Use square-cropped images wherever thumbnails appear in the public interface.'),
             'class' => 'checkbox',
             'value' => 1
         ));
--- a/application/forms/AppearanceSettings.php
+++ b/application/forms/AppearanceSettings.php
@@ -1,7 +1,7 @@
 <?php
 /**
  * Omeka
- * 
+ *
  * @copyright Copyright 2007-2012 Roy Rosenzweig Center for History and New Media
  * @license http://www.gnu.org/licenses/gpl-3.0.txt GNU GPLv3
  */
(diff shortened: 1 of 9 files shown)


application/views/scripts/collections/show.php
```diff
@@ -18,7 +18,8 @@
 
             <?php if (metadata('item', 'has thumbnail')): ?>
             <div class="item-img">
-                <?php echo link_to_item(item_image('square_thumbnail', array('alt' => $itemTitle))); ?>
+                <?php $useSquareThumbnail = (option('use_square_thumbnail')) ? 'square_thumbnail' : 'thumbnail'; ?>
+                <?php echo link_to_item(item_image($useSquareThumbnail, array('alt' => $itemTitle))); ?>
```
zerocrates: Since you changed item_image itself, no need to do another check here, just call item_image with a null argument

application/forms/AppearanceSettings.php
```diff
@@ -18,41 +18,47 @@ public function init()
     {
         parent::init();
 
+        $this->addElement('checkbox', 'use_square_thumbnail', array(
+            'label' => __('Use Square Thumbnails'),
```
zerocrates: This probably needs some sort of description

zerocrates:
I think it will be better to do this check in the FileMarkup view helper, rather than the helper functions where you have it now. Both of those functions just call to image_tag in FileMarkup, so just one change there will accomodate both of them (and all you'd have to change in them is adding the `null` default argument).

The other, more important reason is that we also want to have the switch work for thumbnails that are shown by calls to the `file_markup` helper (these go through the `derivativeImage` method of FileMarkup, which would need a little work to allow null as an image size, and also at `$_callbackOptions` to default to `null` for `imageSize`).

The text of the option and the description I think should stress that this is a _default_ only, because it can be overridden by a theme or exhibit's specific choice of thumbnail.

zerocrates:
So, that change accomplished part of what I was talking about (moving up a level to image_tag) but not the  other part (making this also apply to thumbnails coming from file_markup, rather than only for explicit usages of thumbnails as it is now).

Additionally, and it's more of a peeve, that last change especially made tons of whitespace only changes to the file. I'd appreciate a revert and replacement with a "clean" version of the same change.

PR #765, opened 2016-12-06
Fixed creation of a class name from a mime type.

Daniel-KM:
Hi,

Fixed creation of a class name from a mime type.

Sincerely,

Daniel Berthereau


zerocrates:
`str_ireplace` to `str_replace`: what, you're not worried about the legendary *uppercase* slash??

PR #768, opened 2016-12-08
Allow setting class to li tag in navigation menu

luku:
This is small update, but very handy. 
```php
nav(['uri' => '/', 'label' => 'Home', 'class' => 'home-nav']);
```
produces
```html
<li><a href="/" class="home-nav">Home</a></li>
```
This is OK in most cases, but sometimes is very useful to set class on wrapping `<li>` instead. So my pull req enables:
```php
nav(['uri' => '/', 'label' => 'Home', 'class' => 'home-nav'], null, ['li_class' => true]);
```
which produces
```html
<li class="home-nav"><a href="/">Home</a></li>
```

### DIFF
--- a/application/libraries/globals.php
+++ b/application/libraries/globals.php
@@ -2689,7 +2689,7 @@ function nav(array $navLinks, $name = null, array $args = array())
     }
 
     $menu = get_view()->navigation()->menu(new Omeka_Navigation($navLinks))
-        ->addaddPageClassToLi(!empty($args['li_class']));
+        ->addPageClassToLi(!empty($args['li_class']));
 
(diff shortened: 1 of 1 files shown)


zerocrates:
In this situation, I've currently just done the call on the returned object (since we actually return the Menu object rather than the rendered string):

```php
echo some_nav()->addPageClassToLi(true);
```

Does this suffice? I don't think adding it into the function directly as in this PR actually enables anything beyond the slightly uglier method above.

luku:
Oh yes, this is actually lot better, thanks! I have nothing against closing it. Anyway I didn't like too much option `li_class`

PR #771, opened 2016-12-21
Backport unsaved warning JS from S

zerocrates:
As promised, the unsaved warning implementation backported from S, as an alternative to #722 and #753.

This is my preferred alternative, but I'd be interested to know if there's some downside or corner case I'm not accounting for with it.

### DIFF
--- a/admin/themes/default/javascripts/globals.js
+++ b/admin/themes/default/javascripts/globals.js
@@ -166,6 +166,7 @@ if (!Omeka) {
             formsToCheck.each(function () {
                 var form = $(this);
                 var originalData = form.data('omekaFormOriginalData');
+                var hasFile = false;
                 if (form.data('omekaFormSubmitted')) {
                     return;
                 }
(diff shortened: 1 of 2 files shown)


luku:
It's still in progress, so perhaps it will be solved later, but I can see few details that this approach won't cover:
- impossible to exclude specific fields (for example on `admin/item-types/edit/1` the `add-element-type` radio)
- it will not notice changes related to `<input type="file">`
- it will not correctly detect changes in `<input>` without `name` attribute (applies to my PR as well), example: `admin/appearance/edit-navigation`
- minor = if you change form and try to delete the record, it will still fire warning

But I noticed my attempt doesn't work with tinyMCE correctly (not sure how is #722), which you fixed here!

zerocrates:
On "impossible to exclude" and "input without name," I don't consider those bugs in the implementation here but bugs in those forms: anything with a name should be relevant for submssion/"unsaved" purposes, anything without a name shouldn't be (though this could be a little tricky in a few areas). I'll look at the specific areas you mentioned.

As for file inputs, that's a good point. It should be pretty trivial to just handle them specially, as any filled-in file input is always dirty/unsaved.

Change and delete: by design it complains if _any_ form on the page has changes and it's not that specific form being submitted. I'm open to changing that but I'm not sure of whether we want it to work that way or not.

zerocrates:
If nothing else, I can apply the file input fix back up to S, so there's a positive outcome already right there

luku:
Yes, this is nice trick:) And doesn't require rewriting the serialization function

Issue #764, opened 2016-12-06
Search index is not updated after renaming a tag


luku:
If you try to edit tag name on `admin/tags` page, it doesn't update the search index for tagged items, so you can still find the old tag when searching (global/simple search). I think this should be done in core. But not sure how. The current code in [TagsController](https://github.com/omeka/Omeka/blob/master/application/controllers/TagsController.php#L96) is assigning new value/name directly, so there is no simple way to compare old value against new value in record beforeSave/afterSave callbacks, unless we fetch the old tag in beforeSave, update some boolean flag to let us know that in afterSave callback we should find all tagged items and save them (to trigger the search index update).
So I wonder if we shouldn't alter the code in `TagsController` into:
`$oldTag->setPostData(array('name' => trim($_POST['value'])));`
and then in `models/Tag` afterSave callback add the logic for checking changes in tag name and updating all tagged items. For plugins (i.e. ExhibitBuilder) that are using tags as well, it would require similar logic in `after_save_tag` hook.
Does it sound good, or do you have other ideas?

zerocrates:
Your concern about the number of records needing to be updated is legitimate: it'd probably have to be a job here.

I don't see this as something small enough to fit into 2.5, which I'm trying to wrap up.

I'm somewhat tempted to say that the solution is to just tell people to reindex after renaming tags, but that's obviously not optimal.

PR #772, opened 2016-12-22
Update search index on tag name change.

luku:
This is solution for issue #764 
It also adds possibility to update search index (instead of always creating new from scratch). The update can be accomplished in 2 ways:
1.  using SQL (by selecting or aliasing `record_type`, `record_id` columns)
```php
Zend_Registry::get('bootstrap')->getResource('jobs')
    ->sendLongRunning('Job_SearchTextIndex', array(
        'sql' => "SELECT record_type, record_id FROM table WHERE x=y"
    ));
````
2. using array (maps record types and ids to update)
```php
Zend_Registry::get('bootstrap')->getResource('jobs')
    ->sendLongRunning('Job_SearchTextIndex', array(
        'records' => array(
            'Item' => array(1,2,3,4, ...),
            'Collection' => array(1,2,3,4, ...),
        )
    ));
````

@zerocrates I know you said it is too big update for version 2.5, but I think it's really practical to be able update the search index via core job and it doesn't feel like a dangerous update or complicated one.

### DIFF
--- a/application/controllers/TagsController.php
+++ b/application/controllers/TagsController.php
@@ -32,7 +32,7 @@ class TagsController extends Omeka_Controller_AbstractActionController
      */
     public function browseAction()
     {
-        $params = $this->_getAllParams();
+        $params = $this->getAllParams();
         
         //Check to see whether it will be tags for exhibits or for items
         //Default is Item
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by luku** [comment]
Well, I see another use case for search index update where none of suggested approaches really works; When adding new record types that should be searchable in `admin/settings/edit-search`, it should update/add this new record type into search index. What would be best way then? Using array as previously, but instead of array of ids it would accept Boolean `true` and index all records for those record types:
```php
Zend_Registry::get('bootstrap')->getResource('jobs')
    ->sendLongRunning('Job_SearchTextIndex', array(
        'records' => array(
            'Item' => true,
            'Collection' => true,
        )
    ));
```
?? Any ideas?

luku:
I noticed the search index is not updated on tag delete either, so I added possibility for it as well, but 2 concerns popped up...
1. When deleting, it's not possible to use SQL. Only map of record types and ids is usable. So I wonder if we should change the data type for `args` column into `mediumtext` or `longtext`? 
2. When doing batch-delete of tags (via some plugin), the search index update may be very CPU/memory intensive and not really optimal. Search indexing of same item with 3 tags being deleted will run 3 times, instead of once. Every deleted tag will spawn new parallel process. Finally, even DB table `processes` may get filled quickly and take lot of space.

PR #773, opened 2017-01-09
Fixed the port for base root in bootstrap.

Daniel-KM:
Hi,

My last commit on Omeka 2!

Sincerely,

Daniel Berthereau

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -50,25 +50,28 @@ if ((isset($_SERVER['HTTPS']) && ($_SERVER['HTTPS'] == 'on' || $_SERVER['HTTPS']
     || (isset($_SERVER['HTTP_SCHEME']) && $_SERVER['HTTP_SCHEME'] == 'https')
     || (isset($_SERVER['SERVER_PORT']) && $_SERVER['SERVER_PORT'] == 443)
 ) {
-    $base_root = 'https';
+    $scheme = 'https';
 } else {
-    $base_root = 'http';
+    $scheme = 'http';
 }
 
 // Set the domain.
 if (!isset($_SERVER['HTTP_HOST'])) {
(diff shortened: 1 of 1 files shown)


zerocrates:
Just to make sure I'm reading this correctly, the problem this fixes is an issue with the scheme checks on the old line 69 (as that variable isn't just the scheme)?

Daniel-KM:
Yes, the port was never added to the base_root, because at line 69, base_root couldn't never be "http" or "https".

zerocrates:
There's more to this than meets the eye, it seems: in my testing to confirm the bug, I don't see it. Obviously the code is broken in the way you're suggesting, but here's the strange thing that's happening that's masking the bug: `$_SERVER['HTTP_HOST']` actually contains the port already. Chrome seems to include the port in its Host: request header so it's already there even though the lines that are supposed to add nonstandard ports are not working.

Perhaps this differs from client to client? The problem it brings up is that fixing this bug would actually break the situation I'm seeing by putting the port on there twice. We can just strip off the port from HTTP_HOST if it's there, but this merits a closer look apparently.

zerocrates:
Apparently the Host header spec has always called for the value to include the port whenever it's nonstandard. 

Are you actually seeing a bug from this, or did you just notice the problematic code? Zend's equivalent code just uses the Host header unchanged where it's present, and only tries to calculate the port when its forced to fall back to SERVER_NAME. Since we just hard depend on Host:, I'm beginning to think that the whole block of code that manually handles the port is just cruft.

Daniel-KM:
Hi,

In the true life, it seems that everybody uses http or https standard ports, else somebody would find this buggy code before. So this is just a problematic code that doesn't matter.

Sincerely,

Daniel Berthereau

PR #778, opened 2017-02-08
Explain more, which pages/places are affected

luku:
Hi, feel free to correct the texts, I don't know if *show pages* is really in English, but you can see the point

### DIFF
--- a/application/forms/AppearanceSettings.php
+++ b/application/forms/AppearanceSettings.php
@@ -60,11 +60,13 @@ class Omeka_Form_AppearanceSettings extends Omeka_Form
 
         $this->addElement('checkbox', 'show_empty_elements', array(
             'label' => __('Show Empty Elements'),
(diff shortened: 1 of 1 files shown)


zerocrates:
I think "show pages" is perfectly legit here.

zerocrates:
@mebrett @kimisgold happiness with these changes? (adding the description "Include empty elements in show pages." to Show Empty Elements and "Include element set headings in show pages and in drop-down lists." to Show Element Set Headings in Appearance Settings?

We can tweak the language and/or align with the documentation if necessary. I do believe these are accurate, i.e. that they don't affect anywhere outside of the mentioned locations.

kimisgold:
This all makes sense to me.

mebrett:
Ditto.

zerocrates:
Really should have merged this one a long time ago... let's not look at the dates involved.

Issue #779, opened 2017-02-17
Omeka.warnIfUnsaved - tinyMCE + HTMLPurifier problem


luku:
Warning about unsaved changes appears, even though no changes were made. It applies to all pages where tinyMCE appear and the submitted data are filtered by HTMLPurifier.

Steps to repro:
1. Goto Settings > Security and make sure HTML filtering is enabled (checked) or enable it
2. Create new Item, choose any field you want, check "Use HTML" checkbox and create numlist or bullist
3. Save changes and directly edit same item again
4. Try to refresh page or navigate away - warning appears

I spent some time debugging and the problem is that HTMLPurifier is stripping new line after `<ol>` or `<ul>`, while tinyMCE is always trying to prettify the HTML. So the original text is never same as soon as tinyMCE is loaded and the warning appears.
Example:
- in tinyMCE:
```html
<ol>
<li>one</li>
<li>two</li>
</ol>
```
- in DB (after submit of tinyMCE field):
```html
<ol><li>one</li>
<li>two</li>
</ol>
```

I don't have any good idea how to fix it yet. Changes in purifier or tinymce doesn't feel good at all... Any ideas?

zerocrates:
This no longer seems to be an issue, likely because of the update to TinyMCE in the meantime.

PR #780, opened 2017-02-17
Store always values from tinyMCE rather than from textarea

luku:
Solves #779 
The simplest way seems to be by forcing tinyMCE to spit the original value on initial form load

### DIFF
--- a/admin/themes/default/javascripts/globals.js
+++ b/admin/themes/default/javascripts/globals.js
@@ -151,6 +151,9 @@ if (!Omeka) {
         };
 
         var setOriginalData = function () {
+            if (window.tinyMCE) {
+                tinyMCE.triggerSave();
+            }
             $(this).data('omekaFormOriginalData', $(this).serialize());
         };
 


zerocrates:
At least with TinyMCE 4, this doesn't (or at least doesn't always) have the desired effect.

The triggerSave() can happen before the editor is actually created, causing it to have no effect.

luku:
Hm, yes that's bad... It's not just problem of `<ul>`, `<ol>` tags, generally all HTML that uses more than one formatting style has the same problem.
I tried storing original value from textarea with `BeforeSetContent` event and retrieving it with `GetContent` event when collecting the original value in `Omeka.warnIfUnsaved`, but that way I get always the original value, so no changes are submitted to server. In other words, I don't know how to make sure the original value is returned only to `Omeka.warnIfUnsaved` on form load...

zerocrates:
The only workable option I can think of is to listen to TinyMCE's load event and reset the "original" form data at that point... but you do run the risk there of the user being able to make a change in the time between the page load and the editor load, and we'd no longer detect that as a change... maybe that's too corner-case to worry about though.

luku:
I did more testing (also with plugins adding dynamically stuff info forms) and this is my conclusion:

```js
    // admin/themes/default/items/form.php
    // it could be added directly into `Omeka.wysiwyg()` in globals.js
    Omeka.wysiwyg({
        selector: false,
        forced_root_block: false,
        setup: function (editor) {
            editor.on('init', function (e) {
                jQuery(editor.getElement()).parents('form').trigger('o:form-loaded');
            });
        }
    });

    // admin/themes/default/javascripts/globals.js
    var bouncer;
    var setOriginalData = function () {
        window.clearTimeout(bouncer);
        var $this = $(this);
        bouncer = window.setTimeout(function() {
            if (window.tinyMCE) {
                tinyMCE.triggerSave();
            }
            $this.data('omekaFormOriginalData', $this.serializeArray());
        }, 50); 
    };
```

From my testing the first time `o:form-loaded` is triggered, is before tinymce is even listening for the events, so my first attempt is useless. This is not true when adding more form fields from plugins and triggering `o:form-loaded` from them, but we cannot rely on that.

Triggering the `o:form-loaded` when editor is initialized is safe, but it's highly inefficient. Because it triggers the  `o:form-loaded` on every editor initialization. That's why I added the timeout hack. Does it look good? Or do you have any better ideas?

zerocrates:
Oof, I see why you used it but I _hate_ the debouncing here...

tinymce.init returns a Promise, so we should be able to use that to just fire form-loaded once when all the init work is done (i.e., by chaining a call to `.then()` onto the init call).

luku:
Yeah, using `tinymce.init().then()` would be wonderful, but it's probably planned for future. At least I couldn't make it work. It doesn't fire when all editor instances are initialized, and it doesn't return array of editors. Also their docs about [Promise](https://www.tinymce.com/docs/api/tinymce.util/tinymce.util.promise) are dead. I will see if I can come up with something.

zerocrates:
Hmm, I don't have any trouble with chaining `then` onto the init call.

luku:
Could you please share snippet? I still don't know how to make it work...

zerocrates:
I just have `tinymce.init(/* ... */).then(function () { /* runs after init */ });`

luku:
Hmm, me too, in globals.js. I really wonder what I messed up

luku:
So I reverted all local changes I made, made sure I have up-to-date master, but the result is the same for me... `.then(function(editors) { /*   */ })` runs before the editors are initialized and editors argument is empty (I can see them via `tinymce.editors`). `tinyMCE.triggerSave();`has no effect in `then()`.
And when I listen on `init` event in `setup` callback, it gets fired later than the calls in `then()`.

Tested on recent Chrome, FF. Can you please verify the same results? The basic code is like this:

```js
    // admin/themes/default/items/form.php
    Omeka.wysiwyg({
        selector: false,
        forced_root_block: false,
        setup: function (editor) {
            editor.on('init', function (e) {
                console.log('INIT SINGLE');
            });
        }
    });

    // admin/themes/default/javascripts/globals.js:Omeka.wysiwyg()
    tinymce.init($.extend(initParams, params)).then(function() {
        console.log('INIT ALL', arguments);
        tinyMCE.triggerSave();
        jQuery(tinymce.editors[0].getElement()).parents('form').trigger('o:form-loaded');
    });

```

luku:
All clear... I missed the obvious detail that `selector: false` prevents any editor from loading. On items/add|edit pages the logic for selecting which textarea should have tinymce, comes from elements.js,  and it runs after the form has loaded. So nothing weird with tinymce promise.
However, I don't see any simple way around it. In nutshell: I would alter `Omeka.Elements.enableWysiwyg()` to `toggleClass('tinymce-on')` on textarea and run it before `Omeka.wysiwyg()` with `selector: '.tinymce-on'`.
Does it make sense? Or any elegant ideas?

zerocrates:
Closing, see #779.

PR #782, opened 2017-02-24
Use X-Forwarded headers when deciding whether to set HTTPS 

arthurian:
Hi there,

I'm submitting this PR to fix an issue I encountered where the base URL scheme is not set properly  when SSL is offloaded/terminated at a load balancer. The issue is that instead of setting the scheme to *https* as expected, it sets the scheme to *http*.

Omeka can detect that a client has initiated an HTTPS request that was offloaded to a load balancer by checking the [X-Forwarded](http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/x-forwarded-headers.html) headers.

The solution proposed by this PR is to modify `bootstrap.php` to check the `X-Forwarded-Proto` and `X-Forwarded-Port` header values and set the scheme accordingly. 

@zerocrates Thoughts?

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -49,6 +49,8 @@ define('SCRIPTS_DIR', APP_DIR . '/scripts');
 if ((isset($_SERVER['HTTPS']) && ($_SERVER['HTTPS'] == 'on' || $_SERVER['HTTPS'] === true))
     || (isset($_SERVER['HTTP_SCHEME']) && $_SERVER['HTTP_SCHEME'] == 'https')
     || (isset($_SERVER['SERVER_PORT']) && $_SERVER['SERVER_PORT'] == 443)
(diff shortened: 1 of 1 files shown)


zerocrates:
I've implemented this feature in a slightly different way.

pcfens:
Should we look at ae0b01bd to see the details of how it was implemented, or are there other commits too?

zerocrates:
Ah yeah I meant to mark the commits but as the feature was sort of described by a handful of different PRs i didn't.

That commit you linked is just the part dealing with this interacting with the "force ssl" configuration option. The actual feature is represented by [this merge](https://github.com/omeka/Omeka/commit/51aa8767ad9ad01d145a0df0ee867d556359ce0c): basically it's just checking X-Forwarded-Proto for calculation of WEB_ROOT as the PRs generally did, with the addition of patching Zend's serverUrl helper to also take it into account.

PR #784, opened 2017-02-27
Wrong conditions for Tag filtering by tag type

luku:
The current condition produces something like:
```sql
SELECT tags.* FROM tags
LEFT JOIN records_tags ON (records_tags.tag_id = tags.id AND records_tags.record_type = 'Item')
LEFT JOIN items ON items.id = records_tags.record_id
ORDER BY tags.name ASC
```
So for everybody with sufficient privilege to *showNotPublic* Items, it's impossible to filter only tags for specific record type.

### DIFF
--- a/application/models/Table/Tag.php
+++ b/application/models/Table/Tag.php
@@ -89,8 +89,9 @@ class Table_Tag extends Omeka_Db_Table
         $select->reset(Zend_Db_Select::FROM)
                ->from(array('tags' => $db->Tag), array())
                ->joinLeft(array('records_tags' => $db->RecordsTags),
-                    "records_tags.tag_id = tags.id AND records_tags.record_type = $recordType",
-                    array());
+                    "records_tags.tag_id = tags.id",
+                    array())
+               ->where("records_tags.record_type = $recordType");
 
         //Showing tags related to items
         if ($type == 'Item') {


zerocrates:
Is this actually a problem? The additional join to the item table should only be occuring if the `$type` is "Item".

zerocrates:
Oh I see what you're saying, that the second join there, even only for Items, will pick up other tags where they have some tagging with a record ID that happens to correspond with any item's ID. Yeah?

So the problem condition is that you have to be looking for Item tags only, and have showNotPublic privs, then you'll get some additional tags potentially.

luku:
Yes, because it's left join. If you have tags for both Item and Exhibit, it will always show both, unless there is some additional WHERE condition, which is not by default in admin/tags page. Perhaps change the `include_zero` to `false` would also help...

zerocrates:
Ah wait, I think there's a miscommunication at play here... the "include zero" is designed to be the thing that switches this view from "all tags" to "just tags for this type".

Previously it did in fact just not show tags that didn't have anything tagged with the given type, but people were having problems where they had lots of "dead" tags not tagged to anything that weren't visible, so the Tags page was explicitly changed to by default show all the zero-count tags as well.

What you're currently _supposed_ to get is just accurate counts of the tagged records of the given type, but it's expected that you'd have a bunch of zero-count tags shown if they aren't used by that type.

We might be better served with a UI element or something that uses the "include zero" param.

zerocrates:
In other words, the use of left joins and the condition jammed into the join condition here is intentional.

luku:
Oh, now I get it, thanks. That was really my misunderstanding of the feature. Then this PR is ready to be closed:)
I'm also thinking about rewriting the tags action and view so it's using pagination and some simple search, similar to the one in admin/users view. So now when I know I will try to incorporate this include_zero switch into the view as well. Perhaps adding another option in the Record Type dropdown, like "Item only", 'Exhibit only",... would do the trick?

zerocrates:
The Tags page is just a bit of a mismash, as it's intended to do a couple different things. It both serves as the "tag cloud" (without the cloud part) view for taggable records, where you want it to do what you were expecting here, and also the management panel for tags generally, where you want it to show all the tags.

Maybe we should have an "All" records view as the default here, which whould show counts for any record type and also included the zero-usage tags, then the record-specific views would turn on hiding of zero-taggings tags. An All view can't have the numbers link to anything though.

PR #787, opened 2017-02-28
Admin tags page with pagination and search, closes #784

luku:
So this is my try on how the admin/tags page could look like. 

Few notes:
- `tagType` param was removed and is replaced by `type` param as expected in `Table_Tag::applySearchFilters()`
- option *All* in **Record Types** dropdown displays all tags with counts for all possible record types, while specific record types like *Item* or others will force to display only tags with relations to that type. I hope I understood your last comment @zerocrates in #784 
- all sort criteria can toggle sort dir
- in `TagsController.php` I couldn't understand the meaning behind the `$filter['record'] = $record;` part. Should I remove it, or does it have some magic usage?

### DIFF
--- a/admin/themes/default/css/style.css
+++ b/admin/themes/default/css/style.css
@@ -2209,13 +2209,13 @@ header nav {
         box-shadow: 0px 0px 2px rgba(0,0,0,.1) inset, 1px 1px 2px rgba(0,0,0,.1);
         text-shadow: none;
     }
-    #tags-nav span.current:after {
+    #tags-nav span.current a:after {
         padding-left: 5px;
     }
-    #tags-nav span.current.asc:after {
+    #tags-nav span.current.asc a:after {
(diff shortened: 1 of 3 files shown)


zerocrates:
The `record` param is a bit of a mystery to me as well...

There's a corresponding `record` filter on the Tag table, but it's only usable for providing an actual record object to just get that record's tags. That doesn't really fit with retrieving a param from a request like what's happening here.

That particular little branch there has been essentially untouched for 9 years. It probably dates from a time when the actual record filter took a string or integer. I'd say remove it.

luku:
Ok, I still kept in Table_Tag, just in case someone use it in some plugin and calls it internally

zerocrates:
Yes, the filter itself on the table side is actually used, by the Tag mixin. It's what records use to get their own tags whenever they show them.

PR #789, opened 2017-03-05
Batch-edit redirects

luku:
There's a problem with `$this->_helper->redirector('browse', 'items', null, $params);` and multidimensional `$params` array, such as advanced search criteria, so this pull is fixing it.
Besides, it redirects back to selected items, or to same search params for *batchAll* style of batch-edit. It's practical for direct review of performed changes. So maybe not that useful for `_batchEditAllSave()` action, but definitely for standard batch edit.

### DIFF
--- a/application/controllers/ItemsController.php
+++ b/application/controllers/ItemsController.php
@@ -214,29 +214,30 @@ class ItemsController extends Omeka_Controller_AbstractActionController
 
         $delete = (boolean) $this->_getParam('submit-batch-delete');
 
+        $params = json_decode($this->_getParam('params'), true) ?: array();
+        unset($params['admin']);
+        unset($params['module']);
+        unset($params['controller']);
+        unset($params['action']);
+        unset($params['submit_search']);
+        unset($params['page']);
+
         $batchAll = (boolean) $this->_getParam('batch-all');
         // Process all searched items.
         if ($batchAll) {
-            $params = json_decode($this->_getParam('params'), true) ?: array();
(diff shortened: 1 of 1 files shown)


zerocrates:
What is the problem with multidimensional arrays that this fixes?

luku:
Ok, here comes example:
Let's say you search all items, where DC:Title is not empty. It will produce query like: `search=&advanced[0][element_id]=50&advanced[0][type]=is not empty&range=...`.

Now try to reproduce some error in batch edit, for example delete all items. You will be redirected to URL similar to this one: `items/browse/search//advanced//range//...` As you can see, the whole `advanced` condition got lost. 

In errors.log you then see warning and the problem is in [Zend\Controller\Router\Route\Module.php](https://github.com/omeka/Omeka/blob/master/application/libraries/Zend/Controller/Router/Route/Module.php#L264)

It works fine with multiple one-dimensional queries like `range=0-50&user=1` which produces URL: `items/browse/range/0-50/user/1/...`

zerocrates:
Ah okay, I see. So the heart of the issue is that it tries to use the router to express those params instead of just a query string (as we usually do).

luku:
Yes, in one way. I found about it, when I tried to redirect back to the selected items, which was the actual heart of this pull request:) 
Btw. is there some better way to pass to some redirector method the query params as array? Or perhaps using `url()` instead of manual `http_build_query()`. I'm open for improvements

Issue #613, opened 2013-11-19
Cache-busting and queue_css_file()


ebellempire:
I think it would be pretty useful if `queue_css_file()` and/or `queue_css_url()` allowed theme developers to add a version string for purposes of [cache-busting](http://twosixcode.com/notes/view/simple-cache-busting-for-css-and-js), e.g. "style.css?version=1.1"

PR #790, opened 2017-03-07
Default cache busting for all local CSS/JS files

luku:
This was inspired by #613 and the actual upgrade to Omeka 2.5, where manual clearing of cache in browser was needed.
I believe this should be default feature for all local CSS/JS files. By default using `OMEKA_VERSION`, with possibility to disable it by passing any *empty* value, or adapt it after for example plugin version for plugin specific assets.


zerocrates:
This is probably a good idea. It's one that's come up from time to time and I'm not really sure why we haven't gone ahead with it.

This change is breaking several tests, though, likely ones that depend on testing output for the presence of tags or strings for linked assets.

luku:
Yes, I'm sorry for not running tests before doing pulls... Fixed tests now

PR #792, opened 2017-03-10
Allow zero as legit element text

luku:
There are some cases, where custom elements in Item Type Metada set may need "0" as valid text, that's why this PR.

Another possible update would be check against `strlen(trim($var))`. The current `empty($var)` check doesn't filter out *space*, that's why I made only changes that will allow "0" come in, for now, everything else stays as it was. Do you agree with trimming?


luku:
Any ideas why the PHP 7.x tests failed?

zerocrates:
It looks like Travis has changed their installed version of PHPUnit for those versions to one that doesn't work with Classic.

luku:
Now when phpunit works, shall I push update with trim()?

PR #798, opened 2017-03-31
Change files.size type to bigint

jajm:
This allow to store files bigger than 4GiB


zerocrates:
But what happens when you need to upload something bigger than 18 exabytes??

PR #800, opened 2017-04-05
Icons - font-awesome update to 4.7.0

luku:
There came many new useful icons since the version we're using


zerocrates:
@kimisgold Any objection? I can't think of any reason not to update.

kimisgold:
Go for it.

PR #802, opened 2017-04-20
Update bootstrap.php for https offloading

tarheelcoxn:
In the case of an upstream load balancer (such as haproxy) that handles https while backend application servers talk to the load balancer via http, this change allows backend application server hosts to correctly pass along URLs for resources such as theme CSS when the client connects to the load balancer using https.


zerocrates:
The issue with this PR (and #782), is that our bootstrap code is only one place where the server's base URL is computed. Zend Framework's serverUrl helper also does this, and is used both by their code and ours in several places. My concern is that the helper doesn't respect X-Forwarded* headers, so fixing the issue here introduces an undesirable difference between the two.

What I _have_ done is switched things around so that the JS/CSS/etc assets are loaded with relative rather than absolute paths. This change should substantially reduce the importance of the scheme calculation within bootstrap.php.

We could also just backport the Zend Framework 2+ version of serverUrl, which _does_ check X-Forwarded-Proto.

zerocrates:
I've implemented this feature in a slightly different way.

Issue #803, opened 2017-05-04
Update Recaptcha language in security settings


patrickmj:
Looks like there was a language shift in Recaptcha. Instead of "Private" and "Public" key, as we have it in security settings, they now have "Site" and "Secret" key. We should update our language to match what they use, just for clarity's sake.

zerocrates:
is the site/secret naming used even for old keys? since we're still using the old recaptcha that new keys don't work with

patrickmj:
Not entirely sure. This is based on logging in to my recaptcha account and going through the options there. There's a V2 and an invisible option.

patrickmj:
@zerocrates I think we talked about this, then it also came up in a pull request that made Captcha changes. I think the PR also changed some of the wording, so I'm not sure where this stands right now.

mebrett:
Since we're switching to the new captcha, will the options in Omeka chance to site and secret as well (use for documentation purposes)

patrickmj:
Seems reasonable to make the wording on our end the their end match

PR #804, opened 2017-05-04
Change recaptcha language

patrickmj:
Per #803 , this just changes the language about recaptcha to match what's now on the recaptcha site


zerocrates:
Changing to "Site key" and "secret key" as appropriate is fine. This PR also updated the descriptions to match Google's, which I think will just be confusing. (It talks about things like "use this key in the HTML" which our users don't have to do or worry about).

PR #809, opened 2017-05-29
Fixed overloaded properties for Item and Collection record

### LINKED ISSUE #808: Item - Method named getElements() does not exist
Hi,
I noticed the overloaded method [getElements()](https://github.com/omeka/Omeka/blob/master/application/models/Item.php#L76), when accessing `Item->Elements`, is not defined anywhere. Is it bug or old code that wasn't removed yet?
And if it should be kept, what does it suppose to return? All elements from DC + ITM set, belonging to current item? And indexed by *set name* and then by *element name* in same format as `Item->getItemTypeElements()`? Or any other ideas?

**edit:** Same applies to `Collection` record as well. And there is even another undefined method [getElementText()](https://github.com/omeka/Omeka/blob/master/application/models/Collection.php#L58). Shouldn't it use `getAllElementTexts` method like for `Item` record? 
Not sure if I should open new issue for it, or if we can handle all of them here... 

Thanks,
Luk

luku:
Fixes #808

### DIFF
--- a/application/models/Collection.php
+++ b/application/models/Collection.php
@@ -54,8 +54,7 @@ class Collection extends Omeka_Record_AbstractRecord implements Zend_Acl_Resourc
      * @see Omeka_Record_AbstractRecord::__get
      */
     protected $_related = array(
-        'Elements' => 'getElements',
-        'ElementTexts' => 'getElementText'
+        'ElementTexts' => 'getAllElementTexts'
     );
 
     /**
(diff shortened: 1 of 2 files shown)


zerocrates:
Sorry about the delay: our classic work has focused on 2.5.1 and trying to confirm that peoples' migration issues with 2.5 are resolved.

I'm looking to get back to these pending changes which would be slated for 2.6.

PR #814, opened 2017-06-26
ReCaptcha v2 support

### LINKED ISSUE #730: Update captcha to v2?
Looks like ZF1 uses reCaptcha v1, not v2. Should we try to work around ZF to use the new recaptchas?

I see solutions out there to do this, but we'll run into requirement conflicts -- the one's I've looked at so far require a PHP version higher than Omeka's.

luku:
Fixes #730 

New settings added to decide which ReCaptcha version to use. By default the old v1 is used.

Plugins like *Commenting* would need to add conditional captcha adapter switch.
```php
$this->addElement('captcha', 'captcha',  array(
	'label' => __("Please verify you're a human"),
	'captcha' => array(
		'captcha' => get_option(Omeka_Captcha::VERSION_OPTION) == 'v2' ? 'ReCaptcha2' : 'ReCaptcha',
		'pubkey' => get_option('recaptcha_public_key'),
		'privkey' => get_option('recaptcha_private_key'),
		'ssl' => true //make the connection secure so IE8 doesn't complain. if works, should branch around http: vs https:
	)
));
$this->getElement('captcha')->removeDecorator('ViewHelper');
```

Plugins like *Contribute* will work as before.

*TODO*: I see the old URL to http://recaptcha.net is dead so I replaced it with URL to Google's ReCaptcha. But how about replacing Public/Private Key with Site/Secret Key?

### DIFF
--- a/application/controllers/SettingsController.php
+++ b/application/controllers/SettingsController.php
@@ -75,7 +75,8 @@ class SettingsController extends Omeka_Controller_AbstractActionController
                     'html_purifier_allowed_html_elements',
                     'html_purifier_allowed_html_attributes',
                     Omeka_Captcha::PUBLIC_KEY_OPTION,
-                    Omeka_Captcha::PRIVATE_KEY_OPTION
+                    Omeka_Captcha::PRIVATE_KEY_OPTION,
+                    Omeka_Captcha::VERSION_OPTION
                 );
                 foreach ($form->getValues() as $key => $value) {
                     if (in_array($key, $options)) {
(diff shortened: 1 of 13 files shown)


luku:
Oh, I missed PR #804 about changes in texts. Should I use the same texts as Patrick then?

zerocrates:
I think you're fine to keep the strings as you have them.

luku:
No problems, thanks for working on Omeka S!
You're right. I don't know where I've seen it... When I'm searching now, I can only see that is recommended to generate new key-pair, but v2 is working with v1 keys. I've changed the description.

patrickmj:
Thanks!

PR #815, opened 2017-06-28
Placeholder treated as NULL for item-type/collection search criteria

luku:
When un-selecting item-type or collection in advanced item search (by selecting placeholder *Select Below*) the placeholder is treated as NULL and leads to wrong search results.

Note: Calling only `return;` inside `array_map` is treated as `NULL`, which happens with placeholder for collections. Item-type problem is this: `in_array(null, array('')) => true`. So in both cases `$hasEmpty` var is `true`.

### DIFF
--- a/application/models/Table/Item.php
+++ b/application/models/Table/Item.php
@@ -197,10 +197,11 @@ class Table_Item extends Omeka_Db_Table
             if (is_numeric($collection)) {
                 return (int) $collection;
             }
-            return;
+            return '';
         }, $collections);
 
-        $hasEmpty = in_array(null, $collectionIds);
+        // strict check to skip placeholder (empty string)
+        $hasEmpty = in_array(null, $collectionIds, true);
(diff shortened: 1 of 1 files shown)


zerocrates:
Are you seeing this actually happen, or just inspecting the code? I can't reproduce. (If I'm understanding your report correctly, it would mean a default search would actually be for "in no item set" and "in no collection," and that's not occurring.)
 
Lone empty-string values (as is the case for the "Select Below" placeholder) are filtered out at the applySearchFilters stage: https://github.com/omeka/Omeka/blob/master/application/models/Table/Item.php#L387

luku:
Hmm, you're right, it's not repro in the current master. It happens only after merging my PR with multiple item-type/collection search criteria. Because then the filter key is an array instead of string, so it's passed further. I will include it in the mentioned PR and close this one. Sorry for messing it up

PR #820, opened 2017-09-07
Add pagination for items in collections/show

### LINKED ISSUE #819: collections/show has no pagination but limits the query anyway
The CollectionController showAction limits the items query to the number of records that is set, using the result of `$this->_getBrowseRecordsPerPage()`. However, the show page does not have any pagination controls, and this means that collections with more items than the limit simply leave off a lot of items.

This is the offending line: https://github.com/omeka/Omeka/blob/master/application/controllers/CollectionsController.php#L45
```
        $this->view->items = $this->_helper->db->getTable('Item')->findBy(
            array('collection' => $this->view->collection->id), $this->_getBrowseRecordsPerPage());
```

There are two ways to deal with this in my estimation;
- Remove the limit, changing the call to
```
        $this->view->items = $this->_helper->db->getTable('Item')->findBy(
            array('collection' => $this->view->collection->id));
```
- Add pagination controls to the show page.

In our fork we opted to remove the limit. Let me know which sounds most reasonable and I'll submit a pull request.

zploskey:
Fixes #819.

### DIFF
--- a/application/controllers/CollectionsController.php
+++ b/application/controllers/CollectionsController.php
@@ -41,8 +41,22 @@ class CollectionsController extends Omeka_Controller_AbstractActionController
     public function showAction()
     {
         parent::showAction();
-        $this->view->items = $this->_helper->db->getTable('Item')->findBy(
-            array('collection' => $this->view->collection->id), $this->_getBrowseRecordsPerPage());
+        $currentPage = $this->getParam('page', 1);
+        $recordsPerPage = $this->_getBrowseRecordsPerPage('items');
+        $itemTable = $this->_helper->db->getTable('Item');
+        $params = array('collection' => $this->view->collection->id);
+        $items = $itemTable->findBy($params, $recordsPerPage, $currentPage);
+        $totalRecords = $itemTable->count($params);
+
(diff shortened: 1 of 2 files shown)


zploskey:
Looks like it failed on PHP 5.3 because it could not fetch the PHP 5.3 binary.

zploskey:
Were there any issues with these changes?

zerocrates:
Ah, apologies for not commenting earlier.

There's not really an issue with the change on a technical level per se, it just goes a little bit against the idea of a "show" page.

The collection show view is intended to just give you a preview of the items in the collection, and you'd view the actual items themselves on the normal browse page, just limited to the items in the collection.. That's why the "Items in this collection" heading is a link to the appropriate browse page. Having pagination here kind of turns the collection show page into a much more limited quasi-browse, which I'm not sure we want to do.

We have a separate issue for improving that link and the heading to clarify what's happening on this page, which should be done for Omeka 2.6.

zploskey:
I'd be quite happy as long as it's very clear there are additional items in the collection and an obvious way to get to them. Sounds like that's in the works, so I'll close.

PR #834, opened 2018-01-18
We should not support Captcha V1. It will soon stop working. Users st…

highcalibrewizard:
We should not support Captcha V1. It will soon stop working. Users still on V1 will need to upgrade their keys. V1 keys *might* still work/migrate to V2 after March 31, 2018:
https://groups.google.com/forum/#!topic/recaptcha/Cd0vT0azu-c

V2 is set as default in my code, V1 removed from selection.


patrickmj:
We have similar changes in play in anticipation of the version 1 shutdown. In particular, we've discovered some quirks in Google's API about the keys that we're trying to sort out as best we can.

The upshot is that I think this PR might conflict with what we're working on off the base branch.

We very much appreciate your attention to this. A fix will come out very soon.

PR #857, opened 2018-06-12
Allow plugins to inject HTMLPurifier config/definitions

luku:
As mentioned in [forum](https://forum.omeka.org/t/best-way-to-extend-htmlpurifier-definitions/6399) if one plugin reads/get values or definitions, the object gets finalized so it's impossible to write/set more values to it later. So other plugins must use `$configCopy = HTMLPurifier_Config::inherit($currentConfig);` 
That is also reason why I included the defaults and allowed elements and attributes as filter arguments (no need to get definitions in order to find out what are the current allowed elements)

Feel free to propose different filter name, perhaps shorter: `html_purifier_config` is sufficient?


zerocrates:
Looks good to me, thanks!

PR #859, opened 2018-07-03
Fixed the title of the citation.

Daniel-KM:


zerocrates:
This just flips the comma from within the quotation marks to outside?

That's a regional difference in standard usage. I believe the default citation is supposed to be Chicago or Chicago-esque style, which predictably uses the American-style "inside" punctuation. (I actually agree that this makes no sense but it's the style.)

We could conceivably "translate" this to allow variation but that would go beyond this simple example and at that point the item_citation filter is likely the better option anyway.

Daniel-KM:
Yes, Chicago should be fixed...

PR #860, opened 2018-07-04
Fixed a deprecation notice in Zend libraries for full compatibility with php 7.2.

Daniel-KM:
Because the issue is in the cache, it is used anywhere and it breaks some plugins and, in particular, the queries that return a json.


zerocrates:
Yeah I tried to clean out the `each` es but I guess I missed one... I wonder how old the underlying code is here that it uses that list/each construction? Definitely "vintage."

zerocrates:
Or anyway I remember doing so? Ah well...

PR #875, opened 2019-01-08

ebellempire:


ebellempire:
Should I resubmit this with an @ to suppress the `undefined index` warning in the automated tests?

$force_ssl = (bool) @$this->_options[self::FORCE_SSL];

ebellempire:
Will resubmit, thanks!

zerocrates:
One more thought: since you're doing the comparatively more expensive `parse_url` there maybe it's better to do that calculation in the constructor so it only happens once regardless of how many URLs are generated... so you'd just create the base URL once and save it in a property that would get used in getUri().

It's not terribly significant but there's really no reason to re-do the operation for every URL.

ebellempire:
Resubmitted: https://github.com/omeka/Omeka/pull/876

PR #876, opened 2019-01-08

ebellempire:

### DIFF
--- a/application/libraries/Omeka/Storage/Adapter/ZendS3.php
+++ b/application/libraries/Omeka/Storage/Adapter/ZendS3.php
@@ -22,6 +22,7 @@ class Omeka_Storage_Adapter_ZendS3 implements Omeka_Storage_Adapter_AdapterInter
     const ENDPOINT_OPTION = 'endpoint';
     const BUCKET_OPTION = 'bucket';
     const EXPIRATION_OPTION = 'expiration';
+    const FORCE_SSL = 'forceSSL';
 
     /**
      * @var Zend_Service_Amazon_S3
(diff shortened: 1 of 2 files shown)


application/libraries/Omeka/Storage/Adapter/ZendS3.php
```diff
@@ -63,6 +64,11 @@ public function __construct(array $options = array())
         if (!empty($options[self::ENDPOINT_OPTION])) {
             $this->_s3->setEndpoint($options[self::ENDPOINT_OPTION]);
         }
+        
+        if (isset($this->_options[self::FORCE_SSL]) && parse_url($this->_s3->getEndpoint(), PHP_URL_SCHEME)=='http'){
+            $endpoint = 'https:' . substr($this->_s3->getEndpoint(), 5);
+            $this->_s3->setEndpoint($endpoint);
```
zerocrates: This won't actually work, at least not in your situation: it's the same as just setting an https endpoint in the config directly.

ebellempire: Really? It seems to be working for me.

zerocrates: The URLs would generate correctly but if you were having issues saving and deleting with the endpoint-in-config route before, then you should still be having them with this method.

In the same way, if this code _is_ totally working for you, then specifying the https endpoint directly should _also_ be working. They're totally equivalent.

The distinction with the "workaround" is that it uses a different URL as the base for building URLs to stored files than the one it uses to connect to S3 and make changes.

ebellempire: I'm an idiot. You are correct. I see what I did there and why it doesn't actually work.

PR #877, opened 2019-01-08

ebellempire:

### DIFF
--- a/application/libraries/Omeka/Storage/Adapter/ZendS3.php
+++ b/application/libraries/Omeka/Storage/Adapter/ZendS3.php
@@ -22,6 +22,7 @@ class Omeka_Storage_Adapter_ZendS3 implements Omeka_Storage_Adapter_AdapterInter
     const ENDPOINT_OPTION = 'endpoint';
     const BUCKET_OPTION = 'bucket';
     const EXPIRATION_OPTION = 'expiration';
+    const FORCE_SSL = 'forceSSL';
 
     /**
      * @var Zend_Service_Amazon_S3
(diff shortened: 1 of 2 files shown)


application/libraries/Omeka/Storage/Adapter/ZendS3.php
```diff
@@ -63,6 +64,10 @@ public function __construct(array $options = array())
         if (!empty($options[self::ENDPOINT_OPTION])) {
             $this->_s3->setEndpoint($options[self::ENDPOINT_OPTION]);
         }
+        
+        if (isset($this->_options[self::FORCE_SSL]) && parse_url($this->_s3->getEndpoint(), PHP_URL_SCHEME)=='http'){
```
zerocrates: I was puzzling over the "any value" in the documentation and then noticed this... I was thinking of isset more in the way of a "guard" for checking the actual value, but I think we should still be checking that it's true rather than simply if it's set.

application/libraries/Omeka/Storage/Adapter/ZendS3.php
```diff
@@ -63,6 +64,10 @@ public function __construct(array $options = array())
...
+            $this->_options['FORCE_SSL_DISPLAY_ENDPOINT'] = 'https:' . substr($this->_s3->getEndpoint(), 5);
```
zerocrates: This should probably be a property on the class and not in `_options`, it's clearer if `_options` is just the options set by the user only.

ebellempire: Do you have a preference regarding the endpoint logic? e.g.

```
$endpoint = isset($this->_force_ssl_display_endpoint) ? $this->_force_ssl_display_endpoint : $this->_s3->getEndpoint();
```

versus...

`$endpoint = somePrivateMethod()`;

PR #878, opened 2019-01-09
Add a new default mime type, image/vnd.adobe.photoshop

ZarriLee:
Our customers have JPEG images with mime type “image/vnd.adobe.photoshop”. To allow them to use this mime type as a default we propose the following pull request.


zerocrates:
Do you have an example of a file like that? I'd expect image/vnd.adobe.photoshop to be a .psd not a .jpg.

ZarriLee:
You are right, it seems like they have renamed psd to jpeg =S.

PR #881, opened 2019-02-22
Skip empty arrays when creating IN() condition

luku:
Without this, I need to always check if the array parameter is empty and remove it prior calling `getSelectForFindBy()`.


zerocrates:
My only concern is whether this should be falling down to an "always false" type construction... dev expectations of a query asking "is X in [some set]" might reasonably be that it should be false if that set is empty.

luku:
Not sure I get your point. Empty IN() causes SQL error, so there's no reason to allow unsupported expression. There's another similar problem with NULLs, which I addressed in PR #749, but there I didn't skip empty parameters. So actually neither of my pull requests is ultimate solution:)
I think I will update #749 instead and close this one

PR #905, opened 2020-02-26
fix markup in admin/users/edit page

Mednikov:
Fix incorrect markup that causes problem with user role select. Wrapped a element doesn't let change user role in russian localization.


zerocrates:
Thanks for the report and fix here.

Though the translations are stored in this repository, we actually handle work and updates to them through [Transifex](https://www.transifex.com/omeka/omeka). Any changes made directly here will be overwritten when we update from Transifex, so for translation fixes you should instead sign up and submit them there, not here.

I've gone ahead and applied this fix to the Russian translation in Transifex, so it will get picked up the next time we update the translations (in general, at the release of the next version).

PR #908, opened 2020-03-23
Added freedom to choose field type

DBinaghi:
The type "text" was hardcoded in the <input> field, making it impossible to choose something different, like "password" or "color" f.i.
With this change, it's possible to pass an extra, optional, parameter, letting user choose the type for the field (default: text).
Hope this helps.


zerocrates:
We already have a form element and view helper for this purpose: it's called FormInput, and you pass the type in the `attribs` array.

zerocrates:
Also, "Password" is handled by its own helper/element from Zend Framework, FormPassword.

DBinaghi:
Thanks for your replies, John. I think FormInput will do the job, then. Cheers.

Issue #913, opened 2020-04-09
Minor issue: default fallback thumbnail not overridable


DBinaghi:
Hi.

I've developed a plugin to add extra fallback thumbnails (and types) to the repository, and also replace the original ones, using add_file_fallback_image function (after a suggestion by John, https://forum.omeka.org/t/new-fallbackimage-not-showing-up/7859/4).

Only problem: there's no way to replace the GENERIC_FALLBACK_IMAGE, defined inside the FileMarkup.php.

I know it should be possible to do it via a custom theme, but it'd be nice if the effects of add_file_fallback_image could be extended to the default image too. I suppose  the code in the FileMarkup.php file would have to be changed in order to explicitly assign an image to GENERIC_FALLBACK_IMAGE.

zerocrates:
Yeah, it doesn't make a whole lot of sense to have only the "main" fallback not be modifyable.

Although, unless I'm mistaken, your module should also be able to just provide an alternate image like a theme could, just by placing it at `views/shared/images/fallback-file.png`

DBinaghi:
No, you're not mistaken, that solution works. The problem is that it's not programmatically achieved: if the new fallback_file.png is in that directory, it gets loaded no matter what; with all other images, I can enable or disable them via config form options.
One could say that, if the plugin was installed and enabled, by default the new images were used, though... As I said, minor issue, but if it could be solved for a future release it would be nice.

DBinaghi:
I've tried you solution, but unfortunately it seems not to be working from an Omeka Classic plugin.

DBinaghi:
Found an easy solution. Opened a pull request with it. Hope this helps.

PR #916, opened 2020-04-19
Default fallback image now modifiable via add_file_fallback_image fun…

DBinaghi:
See https://github.com/omeka/Omeka/issues/913

Default fallback image was not programmaticaly modifiable because of it using a constant; I've solved the problem by creating a new entry for the $_fallbackImages array.

### DIFF
--- a/application/views/helpers/FileMarkup.php
+++ b/application/views/helpers/FileMarkup.php
@@ -157,14 +157,8 @@ class Omeka_View_Helper_FileMarkup extends Zend_View_Helper_Abstract
         'audio' => 'fallback-audio.png',
         'image' => 'fallback-image.png',
         'video' => 'fallback-video.png',
+        'generic' => 'fallback-file.png', // Fallback image used when no other fallbacks are appropriate.
     );
-
-    /**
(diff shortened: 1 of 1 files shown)


zerocrates:
This is a pretty reasonable solution to your issue, but I don't love the "generic" key that's being used here... maybe an asterisk `*` would be better?

I also think just for any possible other code that might have referenced it, it's best to leave the `GENERIC_FALLBACK_IMAGE`  constant there.

DBinaghi:
I see no problem in using "*" instead of "generic" (I'd kept "generic" as that was the key used before, I thought for some reasons somebody I had gone for that).
And also in keeping the constant `GENERIC_FALLBACK_IMAGE`, it'd hurt nobody.

zerocrates:
To refer to a class constant within the same class, you need to use `self::`, so `self::GENERIC_FALLBACK_IMAGE`.  The way you have it will cause a warning (and if you're not overriding that entry with a plugin, an error).

The line defining the constant is fine, but as a matter of style, could you move it to the top of the class, above all the static properties?

PR #921, opened 2020-05-02
Fixed numeric html attributes being discarded

php4fan:
I've just encountered this bug, I don't know if there's a bug report about it.

Before this change, `tag_attributes()` would silently discard all attributes with numeric values unless the values were passed as string.
For example:
```
tag_attributes(['width'=>300, 'height'=>200, 'foo'=>'bar'];
// expected result: 'width="300" height="200" foo="bar"'
// actual result before fix: 'foo="bar"'

tag_attributes(['width'=>'300', 'height'=>'200', 'foo'=>'bar'];  // this would already work as expected
```

As a consequence:
```
record_image($someRecord, null, ['width'=>300, 'height'=>200]);
```
would result in an image tag without the expected width and height attributes.

### DIFF
--- a/application/libraries/globals.php
+++ b/application/libraries/globals.php
@@ -1486,6 +1486,7 @@ function tag_attributes($attributes)
         if (preg_match('/[^A-Za-z0-9_:.-]/', $key)) {
             continue;
         }
+        if (is_numeric($attribute)) $attribute = "$attribute";
         if (is_string($attribute)) {
             $attr[$key] = $key . '="' . html_escape($attribute) . '"';
         } elseif ($attribute === true) {


zerocrates:
This looks like a fine change to me...

There's a couple minor matters of code style: we would include the technically-unnecessary curly braces for your one-line `if` (turning it into a three-liner), and I think a cast to string is clearer than the construct you're using here.

Also, a very quick testcase for this would be appreciated (tag_attributes has an existing test class at application/tests/suite/Helpers/TagAttributesTest.php).

php4fan:
Feel free to redo the changes from scratch if it's more convenient than merging this pull request.

It just felt easier to create the PR than to open an issue for what seemed to be a one-liner

PR #930, opened 2020-11-04
Catch general errors during record->save() in re-indexing

timstallmann:
Running into an issue on a current site where the re-indexing process fails out on totally unrelated after-save hooks. It seems to me like this job should be pretty generous in terms of continuing on to process items regardless of whatever exceptions are thrown, so proposing to update this to catch all Exceptions, not just Validation exceptions.

### DIFF
--- a/application/models/Job/SearchTextIndex.php
+++ b/application/models/Job/SearchTextIndex.php
@@ -46,7 +46,7 @@ class Job_SearchTextIndex extends Omeka_Job_AbstractJob
                     // Save the record object, which indexes its search text.
                     try {
                         $recordObject->save();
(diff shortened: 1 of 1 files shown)


zerocrates:
I think you're probably right that this should be pretty aggressive in continuing on errors.

Out of curiosity, what kind of exceptions are you seeing in practice?

timstallmann:
These were storage errors from ArchiveRepertory's `postSave` method, specifically `Omeka_Storage_Exception` due to some filenames being too long (which might be a bug upstream too since I'm not sure where the 190-character filename limit in ArchiveRepertory comes from?

Issue #929, opened 2020-10-28
Input labels for admin edit pages


jaguillette:
Our university IT's accessibility group took a look at Omeka as provided to our community, and identified a few issues. One is that the text area inputs in the admin view don't always get descriptive labels. As they reported to us: 

"In the Admin pages input fields that are text areas are not given accessible names. This has the result of making them indistinguishable from one another to assistive technology. This [screen cast of a screen reader navigating the "Add an Item" page](https://share.getcloudapp.com/GGuRP51Q) shows how every text area is announced as "edit text". There is no way to identify to which section (e.g. Title, Subject, Description) the input corresponds. Note as well in the screen cast the "Use HTML unchecked checkbox" labels. In this case, although the input fields are given accessible names they are not descriptive and do not convey the context of the input."

I'm working on adding input labels for our own Omeka service, and will create a pull request referencing this issue when I have a solution to propose. I'm working off of the new-admin-theme branch so that the changes can be incorporated into the admin theme refresh if desirable.

zerocrates:
This is great. And working off of the new-admin-theme branch is the right move, I think.

How are you currently planning to address this particular issue?

My thinking, when you described it eariler, was to use legends/fieldsets, or possibly aria-role=group and aria-labelledby, to provide the necessary context.

jaguillette:
Adding a `for=` to the existing label would work for the first element, but then subsequent inputs would present a problem, so what seems likely right now is to add a visually hidden label to each textarea. I've done some work for #928 that adds a css class `visually-hidden`, and I'm likely to use the same approach here. The class styling looks like this:

    .visually-hidden {
    	border: 0 !important;
    	clip: rect(1px, 1px, 1px, 1px) !important;
    	-webkit-clip-path: inset(50%) !important;
    		clip-path: inset(50%) !important;
    	height: 1px !important;
    	margin: -1px !important;
    	overflow: hidden !important;
    	padding: 0 !important;
    	position: absolute !important;
    	width: 1px !important;
    	white-space: nowrap !important;
    }

zerocrates:
We've used the group/label solution before, as described by https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA17.html, or the similar https://www.w3.org/WAI/WCAG21/Techniques/html/H71.html that uses legends.

Both don't have the downside of using `for` where the label only applies to one element, but also don't require repeating the group context over again for every element. We'll consider whatever you suggest, but my preference would be one of those two options.

[...]

PR #931, opened 2020-11-16
Input labels for admin edit pages

jaguillette:
This PR is intended to address issue #929 dealing with input labels when editing metadata. It adds a fieldset around metadata inputs and a legend element that's visually hidden but visible to screen readers to group multiple instances of the same metadata input. It also adds an `aria-live: polite tag` around each of the metadata fields, so that screen readers are notified when inputs are added. There's also a minor change to the gray text color to increase the contrast. All of these changes are based on the `new-admin-theme` branch to work with those updates. Happy to revise the changes based on feedback.


jaguillette:
I'm just checking in on this PR in the new year, please let me know if I ought to do things differently to make it easier to merge. Thanks!

zerocrates:
Hi, Jeremy,

I don't think there's anything you need to do, we've just got some moving pieces on our end we need to get into place before we can move on this. Rest assured we haven't forgotten about it.

zerocrates:
Hi, just getting back to this as we're looking at incorporating the new admin theme.

I see you removed the changes to the links to edit, etc. in this PR, which I think is a good idea to keep things separate. Could you also remove the visually_hidden function you added, as now it has no usages in this PR (and I'm not sure a function like that is the direction we'd want to go, actually). It would probably also be good to separate out the gray color change as well into its own issue or PR.

A couple other questions:

- On the labels for the text inputs: does a label that says "text input 1" (etc.) really provide any accessibility? I'm open to having labels there, and if "Text input X" is helpful then we can have that. Other guidance I've seen on multiple inputs seems to indicate not numbering them like this, and just using `aria-labelledby` to point them all at one identical hidden label (for each different element, of course). That would also avoid having to calculate the "for" for each different input. Would that be acceptable/preferable here?
- As an alternative if each input should have a unique label, could we instead use `aria-label`? And thereby avoid the extra elements and linking between them.
- On the legends: would it be preferable to use the existing visible labels as the group labels, either by changing them from `label` elements to `legend` elements, or by using `aria-group` on the field div and connecting that with the existing visible label using `aria-labelledby`?  The setup you have in the PR now of adding an extra hidden legend I assume would lead to some doubled reading. `aria-group` and `aria-labelledby` are easiest for us to work with from a styling perspective (legends and fieldsets are often difficult to accurately position across browsers), so if that's an acceptable approach I think we might prefer it.

jaguillette:
Thanks for that feedback! I'll separate out the color issue and remove the `visually_hidden` function and associated css. I'll try to use aria-labelledby tags instead of hidden elements wherever possible (hopefully everywhere).

As for the question about numbering the text inputs, there are two things that should be able to do. One is to make it explicit which "remove" and "use html" inputs correspond to which text fields. The other is to provide a way for people using voice interfaces to have a way to explicitly reference a given input field. As part of removing the visually-hidden class, I'll take another pass at those inputs and labels. I just learned about chaining aria-labelledby references from our accessibility team, which should provide some elegant solutions.

I'll also give rebasing a try, and defer to you if it proves more than I can handle.

jaguillette:
I've changed things around, hopefully all for the better. I've removed all of the visually-hidden elements, css, and functions, relying entirely on existing elements. I had to add some IDs to elements that didn't have any, as I used chained `aria-labelledby` tags to produce screen-reader friendly labels for all of the inputs. This has an advantage over using groups, whether through a `fieldset` tag or a `role="group"`, since the labels are announced in full by a screen reader, even when navigating through the form elements in isolation, similar to how a screen reader can browse links. With groups, the group is only announced when entering a group by tabbing through inputs, but the inputs remain undifferentiated when a screen reader pulls all of the form inputs into a list for a user.

I was also able to rebase the branch, so it's building on more solid foundations.

zerocrates:
Okay, we've merged a re-rebased version of this (squashed down to a single commit) so it will appear with the new admin theme.

Thanks for your contribution.

Issue #926, opened 2020-10-06
Use `record_url` for View Public Page in Admin items edit screen


timstallmann:
If a site overrides the RecordUrl view helper you'd expect the admin interface "View Public Page" button to pull that URL as well, but currently it is hardcoded in /admin/themes/items/edit.php as /items/show/<item-id>.

This seems like a simple fix and happy to submit a PR for it, let me know!

zerocrates:
Yes, of course feel free to submit a PR (I say months later).

There'll be a slight wrinkle here I think, though: there is no "public" version of record_url that will link from the admin to the public. So you'd have to work around that.

PR #940, opened 2021-03-15
Issue #926: Use record_url for path to public page on admin interface

timstallmann:
For themes/modules that override default URLs, this ensures that the "View Public Page" link on the admin interface goes to the correct URL for an item on that particular site, rather than forcing `/items/show/<id>`

### DIFF
--- a/admin/themes/default/items/edit.php
+++ b/admin/themes/default/items/edit.php
@@ -17,7 +17,7 @@ echo flash();
     <section class="three columns omega">
         <div id="save" class="panel">
             <?php echo $this->formSubmit('submit', __('Save Changes'), array('id'=>'save-changes', 'class'=>'submit big green button')); ?>
-            <a href="<?php echo html_escape(public_url('items/show/'.metadata('item', 'id'))); ?>" class="big blue button" target="_blank"><?php echo __('View Public Page'); ?></a>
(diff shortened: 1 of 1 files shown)


zerocrates:
Have you checked this _without_ having a plugin overriding the URL? I think the problem I mentioned in your initial issue would still happen here.

timstallmann:
Have not tested it yet, but I will!

timstallmann:
@zerocrates just tested finally, and on a default install this just links to `items/show/item_id`, as it should.

timstallmann:
Nope it was the regular `/items/show` page. I mean, please do test it yourself too to verify, I'd be happy to work on the code further if need be.

timstallmann:
Huh strange -- let me see if I can replicate that on my end

zerocrates:
I alluded to this problem in your original issue #926: links that go from one "side" to the other need special handling. For links created from strings, we have `public_url` and `admin_url` functions that force the URL to be created on the correct side.

I think the answer here involves creating new `public_record_url` and `admin_record_url` functions and using the public one here. If you look at `public_url` and `admin_url` you'll see that the process involved is pretty simple.

timstallmann:
Got it, thanks! Will submit a revised PR next time I can squeeze in some time..

PR #965, opened 2021-09-21
Allow linking to records of any type from Browse Tags page.

kloor:
Our librarians requested that they be able to navigate to records besides Items from the Browse Tags page. This attempts to do so in a generic way, so that as long as you are filtering to a specific Record Type, the number next to a tag links to all records of that type with that tag. This works for Exhibits, as well as Collections as I proposed in #963.


DBinaghi:
Seems like a good idea to me. Tags browing page needs some improvement and restyling (see https://github.com/omeka/Omeka/issues/993)

PR #973, opened 2022-02-19
Use supported PHP in GHA

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>

https://www.php.net/supported-versions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


zerocrates:
We're not going to cut out the post-EOL PHP versions from the matrix here: they're still out in the wild in many places with distribution-provided security support. Increases to the minimum required PHP version for Classic would be based around necessary functionality, or if they became particularly onerous to support.

On the increasing side, work to enable PHP 8.1 support will be forthcoming, and we'll enable 8.1 in the testing matrix at that point.

PR #975, opened 2022-03-12
fix typo in application/languages/ja.po

tsunekawa:


zerocrates:
Hello,

The translations in those files come from the crowdsourced translation platform Transifex. We don't make changes to the translations directly in this repository, as they would be overwritten by the updates from Transifex.

You can sign up to the [Omeka project on Transifex](https://www.transifex.com/omeka/omeka/) and make your fix there. Then it will be picked up the next time we pull in the updated translations.

tsunekawa:
Hello,

Thank you for sharing about how to translate. I will modify the translation using Transifex.

PR #977, opened 2022-06-07
Use bcrypt for passwords instead of SHA1

dicksonlaw583:
This changes the password hashing method to bcrypt, which is much more resistant to mass computation and now the industry standard for storing passwords.

### DIFF
--- a/application/libraries/Omeka/Auth/Adapter/UserTable.php
+++ b/application/libraries/Omeka/Auth/Adapter/UserTable.php
@@ -42,7 +42,9 @@ class Omeka_Auth_Adapter_UserTable extends Zend_Auth_Adapter_DbTable
         $resultIdentities = parent::_authenticateQuerySelect($dbSelect);
         $correctResult = array();
         foreach ($resultIdentities as $identity) {
-            if (password_verify($this->_credential, $identity['password'])) {
+            if ((version_compare(PHP_VERSION, '5.5.0') >= 0) ? 
+                password_verify($this->_credential, $identity['password']) :
+                (crypt($this->_credential, substr($identity['password'], 0, 29).'$') == $identity['password'])) {
                 $identity['zend_auth_credential_match'] = 1;
                 $correctResult[] = $identity;
             }
--- a/application/libraries/Omeka/Auth/Adapter/UserTable.php
+++ b/application/libraries/Omeka/Auth/Adapter/UserTable.php
@@ -42,9 +42,7 @@ class Omeka_Auth_Adapter_UserTable extends Zend_Auth_Adapter_DbTable
         $resultIdentities = parent::_authenticateQuerySelect($dbSelect);
         $correctResult = array();
         foreach ($resultIdentities as $identity) {
-            if ((version_compare(PHP_VERSION, '5.5.0') >= 0) ? 
-                password_verify($this->_credential, $identity['password']) :
-                (crypt($this->_credential, substr($identity['password'], 0, 29).'$') == $identity['password'])) {
+            if (password_verify($this->_credential, $identity['password'])) {
                 $identity['zend_auth_credential_match'] = 1;
                 $correctResult[] = $identity;
             }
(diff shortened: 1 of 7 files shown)


zerocrates:
Replacing the home-grown SHA-1 hashing/salting with password_hash is definitely a good idea.

There's at first certainly the wrinkle of this requiring PHP 5.5, though obviously extensive old-version support is less relevant these days than it used to be. But it probably informs us on what kind of release we'd put this in minimally. We could also use ircmaxell's password_compat library to extend compatibility to our current 5.4 lower limit... part of the reasoning behind choosing 5.4 when increasing previously was to both ensure Composer compatibility and get beyond the early 5.3 versions that had a problem with their bcrypt implementation.

As for how it actually is implemented, it looks pretty good to me. Reusing the salt column with just the content "bcrypt" is a clever solution to being able to tell the difference between the formats for "free" without having to add another column.

In terms of some specific nits:

I think I'd basically reverse the order of your checks when seeing if you need to "reset" the password: check for `$salt === 'bcrypt'` first as it would become the usual expectation going forward, then `$salt !== null` to check the previous format, then just the hash check for the ancient format. This also results in fewer redundant checks against what the salt is or isn't than the order you're doing it now.

Along those same lines, we might as well build in password_needs_rehash support while doing this at the same time. And it then probably makes sense to just use PASSWORD_DEFAULT instead of PASSWORD_BCRYPT then, and just bite the bullet and use a 255-character hash column as the PHP manual suggests (and we could go ahead and collate it as latin1_bin while we're at it since we really don't need UTF-8 there).

zerocrates:
On your last commit, I think just leave to us whether we want to just bump the version to 5.5 or go through the trouble of pulling in the password_compat polyfill. It does various other things to try to get "better" randomness for the salt that are worth getting vs. the mt_rand thing you're doing there.

So you could revert that. I don't think you need to actually make any changes to accommodate my concern about the PHP verisons.

zerocrates:
What do you think about my suggestion about PASSWORD_DEFAULT and the larger column size?

dicksonlaw583:
I have some doubts about `PASSWORD_DEFAULT` as [the description on the PHP Manual](https://www.php.net/manual/en/function.password-hash.php) says that its format and algorithm can change over time. Using that would make it difficult for us to stay consistent across different PHP versions going forward. With `PASSWORD_BCRYPT`, we know exactly what to expect out of it. But having a larger column size is definitely doable for future-proofing purposes.

zerocrates:
Hmm... I suppose in a future version if they changed PASSWORD_DEFAULT it could cause compatibility issues if you ran an installation for a while on that newer version then wanted to move back to an older one which didn't have support for the algorithm. How realistic that is of a concern, I'm not sure (one imagines only an algo that has been supported for quite a long time would ever be switched to be the new default).

Worth noting that we're already using PASSWORD_DEFAULT for Omeka S... but also the difference is currently academic as there's never been a different default so far.

dicksonlaw583:
@zerocrates 

With PHP 5.4 and PHP 5.5 dropped for 3.1, I think now would be a good time to revisit this PR as the bcrypt compatibility issue won't apply going forward. I have just rebased the PR branch awaiting your review. What do you think?

zerocrates:
Thanks for continuing to work on this; I'll take a look soon.

I agree with you that the higher minimum PHP version of 3.1 makes the decisions here easier.

zerocrates:
OK. Use your own judgment on whether this counts as "soon" since my last message, but: this is squashed and merged. 
Thanks for sticking with it.

Rather than drag things out with more back-and-forth I went ahead and made changes on top of the last state of what you had here and that's together what's gone in. The commit has a short explanation of the changes, but discussed here again:

- The migration moving from DEFAULT NULL to NOT NULL caused problems if there were any NULL rows already present; I changed it to keep DEFAULT NULL
- As I'd mentioned as a possibility above, I also migrate to ascii_bin collation for the hash column
- I moved the migration ID to a more current date, just reflecting the time it actually went into the main branch
- I switched to use PASSWORD_DEFAULT instead of PASSWORD_BCRYPT; I know we talked about this and you had some reservations... at the end of the day consistency with Omeka S won out for me, as well as the continuing history of PHP not changing it yet and their assurances of at least one "main" release of adding support for any hash before it becoming the default (and likely significantly more)
- I made some changes to the password upgrade checks to remove some duplicate checks handled by the previous returns
- I added a call to upgrade passwords to the validator (e.g. when resetting a password); if not done here then that check would fail for users that need the hash upgrade until they've actually had to log in again
- I added a couple extra tests for hash upgrades with wrong passwords

PR #989, opened 2022-10-12
Include lightgallery media viewer in the core.

kimisgold:
Provides files and view helper themes can use to display media files in lightgallery. The view helper allows for output that displays either the lightgallery viewer or a list of files that cannot be displayed within lightgallery. This should support images, html5 video, youtube, and pdfs.


application/views/helpers/Lightgallery.php
```diff
@@ -0,0 +1,114 @@
...
+                        'attributes' => [
+                            'preload' => false, 
+                            'playsinline' => true, 
+                            'controls' => true,
+                        ],
+                    ];
+                    if (isset($index['tracks'])) {
+                        foreach ($index['tracks'] as $key => $track) {
+                            $label = metadata($track, 'display_title');
+                            $srclang = (metadata($track, 'dcterms:language')) ? metadata($track, 'dcterms:language') : '';
```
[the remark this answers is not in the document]
zerocrates: also, for this stuff that's going into JSON, you probably want to use `metadata`'s `no_escape` flag or you'll get double escaping (probably)

application/views/helpers/Lightgallery.php
```diff
@@ -0,0 +1,114 @@
...
+                    $html .=  '<div data-src="' . $source . '" ' . $mediaCaptionAttribute . 'data-thumb="' . html_escape(metadata($media, 'thumbnail_uri')) . '" data-download-url="' . $source . '" class="media resource">';
+                    $html .= file_markup($media);
+                }
+                $html .= '</div>';
+            }
+            $html .= '</div>';
+        } elseif (isset($sortedMedia['otherMedia'])) {
+            $otherMedia = $sortedMedia['otherMedia'];
+            $html .= '<div id="other-media" class="element">';
+            $html .= '<h3>' . get_view()->translate('Other Media') . '</h3>';
```
zerocrates, line 71: Inside a helper there's a `$this->view` to get the view the helper was called on: we should be using that instead of the global function here.

application/views/helpers/Lightgallery.php
```diff
@@ -0,0 +1,114 @@
...
+            }
+            $html .= '</div>';
+        }
+
+        return $html;
+    }
+
+    protected function _prepareLightgalleryFiles($files = null) {
+        $sortedMedia = [];
+        $whitelist = ['image/bmp', 'image/gif', 'image/jpeg', 'image/png', 'image/svg+xml', 'video/flv', 'video/x-flv', 'video/mp4', 'video/m4v',
```
zerocrates, line 83: This ultimately sources in part from the HTML5 Media plugin, right? Not sure that the flv ones will really work here (or actually that they still work in HTML5 Media in the absence of Flash). Have you tested it?

Similarly WMV, though maybe Windows machines can play that through a video tag? And I'd actually expect it to be taking Ogg video also, though realistically MP4 under its various names and to a lesser extent WebM are all that are relevant. For reference the core renders as HTML5 video tags the following types:

`video/mp4`, `video/x-m4v`,  `video/ogg`, `video/webm`, `video/quicktime`

zerocrates:
Some remaining questions:

- thumbnails using data-thumb are probably not ever working, `exThumbImage: 'data-thumb'` needs to be passed to lightGallery init for that to work
  - if this _does_ work, do we need the img inside the video div? or do we prefer to keep it so there's some content in there?
  - thumbnails for videos have code to check if the file has thumbnails and to use the fallback, other types don't (probably an unnoticed problem since those data-thumb attrs are dormant)
  - on the previous: it's annoying to get a thumbnail _url_ for an item, instead of a thumbnail _image tag_ which we have a helper for. should we just provide a record_image_url function?
 - Other Files seems a little underbaked: links go always to file _pages_, no thumbnails rendered. standard built-in themes just call `file_markup` on the array of files to display them, though I assume we don't want to do that either

kimisgold:
> on the previous: it's annoying to get a thumbnail url for an file, instead of a thumbnail image tag which we have a helper for. should we just provide a record_image_url function?

I'd love record_image_url(), I'm pretty sure I've run into many other instances of not needing the full image markup, especially with third party jquery plugins.

PR #990, opened 2022-10-25
Improve labeling for "Narrow by specific fields" on advanced search

kimisgold:
This provides the following accessibility improvements for the "Narrow by specific fields" group on the advanced item search form:

* More detailed aria labels
* Visible labels for the individual selects and text inputs
* Alerts to screen readers when field rows have been added or removed

These updates are only implemented for the public views at the moment. I was thinking we could mirror the changes for the admin side once this is approved and merged, though open to doing that within this pull request as well.

### DIFF
--- a/application/views/scripts/items/search-form.php
+++ b/application/views/scripts/items/search-form.php
@@ -21,8 +21,15 @@ $formAttributes['method'] = 'GET';
         ?>
         </div>
     </div>
+    <div id="search-narrow-by-field-alerts" class="sr-only alerts" aria-atomic="true" aria-live="polite">
+        <p><?php echo __('Number of rows in "%s":', __('Narrow by Specific Fields')); ?> <span class="count">1</span></p>
+    </div>
     <div id="search-narrow-by-fields" class="field">
-        <div class="label"><?php echo __('Narrow by Specific Fields'); ?></div>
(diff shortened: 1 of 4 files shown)


application/views/scripts/items/search-form.php
```diff
@@ -34,71 +42,86 @@
...
-                    array(
-                        'size' => '20',
-                        'title' => __("Search Terms"),
-                        'id' => null,
-                        'class' => 'advanced-search-terms'
-                    )
-                );
-                ?>
-                <button type="button" class="remove_search" disabled="disabled" style="display: none;"><?php echo __('Remove field'); ?></button>
+            <?php $index = $i + 1; // The index must start at 1 for accessible labels. ;?>
```
zerocrates, line 52: Seems odd to need to change the indexes here for this to work; all else equal I'd like to keep these zero-indexed.

zerocrates: I see now that actually this doesn't touch $i and instead makes a new $index.... this is just so that it will read "Row 1" to start and not "Row 0"? Might simply prefer to have the relevant code use $i as-is and just add 1 to $i when you're printing the `Row %s` aria-label.

Looks to me that all the search-row-$index id and labelledby stuff would work fine zero-indexed, so just using $i.

kimisgold: Ready for another look.

zerocrates:
Is the public.css that this adds being used anywhere?

PR #1013, opened 2023-11-27
Minor deprecation warning fix in Zend_Ldap.

kloor:
This fixes a minor deprecation warning caused by Zend_Ldap attempting to str_replace() on null, for plugins adding LDAP support to Omeka Classic via the Zend libraries.


zerocrates:
`str_replace("\0", '', (string) $password)` should be equivalent (just adding the cast to string): if that works fine for you that's the form of this fix I'd prefer.

kloor:
That's fine with me, do you want me to update the PR?

zerocrates:
That would be convenient, yeah.

zerocrates:
Thanks.

PR #1014, opened 2023-12-04
Fix PHP 8.2 deprecation warnings for dynamic properties.

kloor:
PHP 8.2 has deprecated dynamically created object properties. These two instances were found while using the insert_files_for_item global helper using the Url  transfer strategy. I've resolved the issue by simply adding the properties. I set the visibility to protected due to the names having leading underscores, but I could see setting them to public to ensure that there would be no broken code. I could not find any instance of the properties being used by any other code within Omeka, but I suppose there is a very slight chance a plugin could expect the side effect of the properties being created.


zerocrates:
I've sometimes handled these with public declarations for the reason you state... here I think I agree that there's probably not an issue with having these be protected.

PR #1015, opened 2024-01-03
Add iframe allowfullscreen attribute to HTML purifier

### LINKED ISSUE #942: Add Iframe allowfullscreen attribute
`*.allowfullscreen` is being stripped by htmlpurifier even when its added in  **Admin -> Settings -> Allowed HTML Attributes**

Debugging reveals that, htmlpurifier throws the following warning message:
`"Attribute 'allowfullscreen' in element 'iframe' not supported (for information on implementing this, see the support forums)"`

This behaviour happens because "allowfullscreen" is not defined in `application/libraries/htmlpurifier/HTMLpurifier/HTMLModule/iframe.php`

Solution / Workaround (I added them in `application/libraries/Omeka/Filter/HtmlPurifier.php/createHtmlPurifier`):

```
$htmlDefinition = $purifierConfig->getHTMLDefinition(true);
$htmlDefinition->addAttribute('iframe', 'allowfullscreen', 'Bool#allowfullscreen');
```

Reference:
[htmlpurifier custom attributes](https://stackoverflow.com/questions/4464004/htmlpurifier-custom-attributes)

dicksonlaw583:
This prevents the `allowfullscreen` attribute from being erased in `<iframe>` tags, which is needed for many types of embedded viewers and players.

Resolves #942

### DIFF
--- a/application/libraries/htmlpurifier/HTMLPurifier/HTMLModule/Iframe.php
+++ b/application/libraries/htmlpurifier/HTMLPurifier/HTMLModule/Iframe.php
@@ -43,6 +43,7 @@ class HTMLPurifier_HTMLModule_Iframe extends HTMLPurifier_HTMLModule
                 'longdesc' => 'URI',
                 'marginheight' => 'Pixels',
                 'marginwidth' => 'Pixels',
+                'allowfullscreen' => 'Bool',
             )
         );
     }


zerocrates:
That's an upstream library that gets updated, so we don't want to have changes directly to it.

There's options for us to allow the attribute without modifying the library. You might consider making this suggestion/contribution more in the form you have it here to [HTMLPurifier itself](https://github.com/ezyang/htmlpurifier) though.

dicksonlaw583:
@zerocrates:

I tried allowing the attribute by adding `iframe.allowfullscreen` or `*.allowfullscreen` to Admin > Settings > Security > Allowed HTML Attributes, but even then the attribute keeps being erased after saving. Please see #942 for details. Though that ticket came from April 2021, I am still able to replicate it in 3.1.2. Do you know of any other ways to work around this?

zerocrates:
So the change you're making, or something like it, is what needs to happen: if HTMLPurifier doesn't know about an attribute, you can't add it using the security settings.

What I'm saying is, we don't want to go editing our copy of HTMLPurifier as it complicates future upgrades to newer versions of the library. Getting HTMLPurifier itself to support the attribute is one way around that problem, as we'd then just pull in the updated version from them. Their legacy support for `object` has a flag for allowing fullscreen, so I could imagine they'd accept something similar for iframes.

Now, staying within Omeka itself, there's still an option for varying the allowed attributes without modifying the library. Our code that uses the purifier lives in Omeka_Filter_HtmlPurifier, and you can use [HTMLPurifier's API for adding an attribute](http://htmlpurifier.org/docs/enduser-customize.html) there.

dicksonlaw583:
I see. While working on the issue last week, I settled on creating a quick plugin for it to solve my use case: [iFrame Allow Fullscreen](https://github.com/dicksonlaw583/IframeAllowFullscreen)

Since this is an upstream dependency not related to Omeka's core, I will close this PR for now.

PR #1017, opened 2024-01-20
Multiple file input

jimsafley:


zerocrates:
Test failures on 7.2 and lower are due to the heredoc: indentation of the closing line is a 7.3+ feature, before then it has to be flush left (also the leading whitespace in the body won't be stripped).

PR #1021, opened 2024-03-05
Update uk.po

vovanshu:
Added and update strings translating


zerocrates:
Translations are handled through a separate service, Transifex, so we don't accept changes to them directly here.

You can view and join the Omeka S Transifex project at https://explore.transifex.com/omeka/omeka-s/

PR #1023, opened 2024-03-05
Update Item.php

vovanshu:
Fix line punctuation


zerocrates:
This is a difference of conventions: standard American style is for commas to go inside quotation marks.

Note that you can use the item_citation filter to change the citation to whatever style you want.

PR #1026, opened 2024-03-06
Offer to set the option to sign an e-mail from whom

vovanshu:
Offer for update GeneralSettings.php
Added the option to sign an e-mail from whom.
Offer for update UsersController.php
Added a condition to use an email signature from who if the option is set.

The reason for this proposal was the problem of a broken signature of an e-mail from whom.
Maybe it will be useful to someone, because it turned out to be necessary for us due to the fact that the name of the site cannot be shortened.

When using a long signature in Cyrillic from the site title.

msg1

When using a short signature from a custom field.

msg2

When using a long signature in Latin from the site title.

msg3

### DIFF
--- a/application/controllers/UsersController.php
+++ b/application/controllers/UsersController.php
@@ -98,6 +98,8 @@ class UsersController extends Omeka_Controller_AbstractActionController
     private function _sendResetPasswordEmail($toEmail, $activationCode)
     {
         $siteTitle = get_option('site_title');
+        $emailFrom = get_option('email_from');
+        $from = get_option('administrator_email');
 
         $mail = new Zend_Mail('UTF-8');
         $mail->addTo($toEmail);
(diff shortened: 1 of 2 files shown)


zerocrates:
We've seen an error before that's similar to this but not quite the same.

Can you try a different change for me, keeping your site's original, longer name:

https://github.com/omeka/Omeka/commit/ccd001f85d14b9b1c13e96cdb8f963aabdf6f9b8

This changes the newlines used in longer header values like your problematic From header, and may resolve this problem.

vovanshu:
I tried the way you suggested to solve our problem, with the signature of the e-mail from whom. But unfortunately, the problem remained, most of all the problem is related to the use of Cyrillic and the long title of the site.
PS: Previously I restored the original code without my changes.

screenshot4

screenshot5

zerocrates:
Good to hear that mail.force_from/mail.force_from_name are working for you. I think we'll probably keep with that rather than your change here then.

On your tests with different names: just to make sure I'm reading right,

It looks like neither 1 or 2 worked, 3, 4 and 5 are fine, and it looks like 7 worked also?

PR #995, opened 2022-11-03
Standardize drawer function for consistency and accessibility

kimisgold:
This pull request consolidates multiple uses of the "drawer" pattern where a component can collapse and expand, and often be marked for removal or be dragged to create order. It also does the following:

* Many interactive elements are now buttons, where they had been previously anchors or divs.
* Buttons and other form elements now have descriptive aria-labels and titles.
* There are now screen reader alerts for when new drawers have been added to the drawer list.
* There is now `Omeka.manageDrawers()` in `globals.js`, to centralize near identical functionality spread across js for navigation and item type elements.

PR #1032, opened 2024-07-10
Drawer toggle a11y rebase

kimisgold:
Your rebase seems to keep everything from the original branch (#995). I'll include the original description here too.

---

This pull request consolidates multiple uses of the "drawer" pattern where a component can collapse and expand, and often be marked for removal or be dragged to create order. It also does the following:

* Many interactive elements are now buttons, where they had been previously anchors or divs. 
* Buttons and other form elements now have descriptive aria-labels and titles.
* There are now screen reader alerts for when new drawers have been added to the drawer list.
* There is now Omeka.manageDrawers() in globals.js, to centralize near identical functionality spread across js for navigation and item type elements.


zerocrates:
Notes:

- should combine "Undo" and "Remove" into a single string for translation
- "Remove" doesn't seem to work correctly on the item type form: for both adding new and existing elements, clicking "remove" will still end up with the element getting added; only the order input looks to be affected, so the elements will come in at the top rather than wherever they were supposed to be positioned before being "removed"

kimisgold:
This branch has been tested against [ExhibitBuilder](https://github.com/omeka/plugin-ExhibitBuilder/issues/143), [VraCore](https://github.com/omeka/plugin-VraCore/issues/111), [ItemOrder](https://github.com/omeka/plugin-ItemOrder/issues/10), and [Contributor](https://github.com/omeka/plugin-Contribution/issues/76). They're all looking good, though VraCore has some screen reader alert quirks that I think will be resolved once this branch is merged.

PR #1036, opened 2024-08-29
Misaligned element texts in admin theme

ebellempire:
**NOTE**: I did not compile the updated Sass file because I don't have that set up right now. Instead I just manually added new styles to `style.css`. But the necessary change is small and easy to implement outside this PR if that's preferred.

The screenshot below is an example of what this change addresses. Basically, all element texts after the first one are misaligned when the viewport is smaller than 1440px. I think it's been like this for quite a while.

Screenshot 2024-08-29 at 10 38 15 AM

### DIFF
--- a/admin/themes/default/css/sass/pages/_items.scss
+++ b/admin/themes/default/css/sass/pages/_items.scss
@@ -241,6 +241,10 @@
         height: 0px;
         margin-bottom: 0;
     }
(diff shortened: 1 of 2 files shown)


kimisgold:
Thanks for this! I pulled your change and compiled the sass.

PR #1037, opened 2024-09-18
Provide alt text configuration options for images.

kimisgold:
* Within the Appearance Settings, users can select an element that can provide alt text for files.
* Users can also write alt text via a new field on the file edit form. This field will override alt text provided by file metadata.
* Images use a blank alt by default.
* File markup no longer provides titles by default, nor populate titles using the alt attribute.
* Browse views no longer assign "presentation" roles to the record images.

### DIFF
--- a/application/models/File.php
+++ b/application/models/File.php
@@ -156,6 +156,8 @@ class File extends Omeka_Record_AbstractRecord implements Zend_Acl_Resource_Inte
                 return $this->getDisplayTitle($this->original_filename);
             case 'rich_title':
                 return $this->getRichTitle(html_escape($this->original_filename));
+            case 'alt_text':
+                return $this->getAltText();
             default:
                 return parent::getProperty($property);
         }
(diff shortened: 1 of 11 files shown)


application/models/File.php
```diff
@@ -154,6 +161,20 @@ public function getProperty($property)
...
 
+    public function getAltText() {
+        $fileCustomAltText = $this->alt_text;
+        $fileAltTextElement = get_option('file_alt_text_element');
+
+        if ($fileCustomAltText) {
+            return $fileCustomAltText;
+        } elseif ($fileAltTextElement) {
+            $fileAltTextElementText = metadata($this, explode(',',$fileAltTextElement));
+            return strip_formatting($fileAltTextElementText);
```
zerocrates, line 172: In general for all this, especially if you change the option to use get_table_options, you might take inspiration from the ElementText mixin's getDisplayTitle. It has some fluff you won't need (like it allows for some HTML tags) but the basic structure of getting the text, checking for the html flag, stripping, decoding, is probably what you'd want/need to properly support possible HTML values here.

If you're in here with an element ID only as the option value, you can use getElementById and getElementTextsByRecord to get the text record you'd need.

zerocrates (commented):
Looks pretty good.

My most substantive comments relate to the fallback to an element and will be found on the settings and in the File model.

PR #1039, opened 2024-11-21
Replace required asterisk with text label.

kimisgold:
The current means of marking required fields is an asterisk, often applied as a pseudo element. It is inaccessible to screen readers, so this pull request replaces the asterisk with a styled text label that reads "required".


zerocrates:
This had one issue which was that you had to turn off escaping to get the extra label markup in there, with the side effect that now the label itself won't be escaped.

I think the way we'd have to go with this is to make our own Label decorator that can include this markup.

kimisgold:
I extended the default label decorator, and overwrote some of the functions so that the label contents are escaped by default or if the option is set to true. First `getLabel()` no longer adds the required prefix or suffix, just retrieves the label contents. In `render()`, the label contents can be escaped before having the non-escaped required content attached using `handleRequired()`. The label contents and required markup then get passed to `formLabel()` with escaping enabled.

PR #1040, opened 2024-11-27
Prevent undefined array key access in FormInput helper

csidirop:
Hi, I wrote some plugins the past days and used the `formInput()` function in my config forms. All occurrences throwing following warning:
`[php:warn] [pid 69748] [client 172.24.0.1:55254] PHP Warning:  Undefined array key "type" in /app/application/views/helpers/FormInput.php on line 42, referer: http://localhost:8202/nikephoros/admin/plugins
`

If we call `formInput()` without an attribute parameter it defaults it to `null`. But later in L42 we accessing that attribute without checking if its set:
https://github.com/omeka/Omeka/blob/56c7f087f2311f8f23adbfac6e7ac7367fb308c2/application/views/helpers/FormInput.php#L41-L45

This PR adds a simple 'isset()' check, so we do not access a unset variable.

### DIFF
--- a/application/views/helpers/FormInput.php
+++ b/application/views/helpers/FormInput.php
@@ -39,7 +39,7 @@ class Omeka_View_Helper_FormInput extends Zend_View_Helper_FormElement
         }
 
         $type = 'text';
-        if ($attribs['type']) {
+        if (isset($attribs['type'])) {
             $type = $attribs['type'];
             unset($attribs['type']);
         }


csidirop:
In addition here is one example how I used it: 

```php
<?php
    $option = get_option('option');
?>

<div class="field">
    <div class="two columns alpha">
        <?php echo get_view()->formLabel('option', __('Option Name')); ?>
    </div>
    <div class="inputs five columns omega">
        <p class="explanation">
            <?php echo __('The Desc.'); ?>
        </p>
        <?php echo get_view()->formInput('option', $option); ?>
    </div>
</div>
```

I didn't found any other plugin which used this, so I could not verify how others used it. So added my code for reference.

zerocrates:
The reason this isn't accounted for, and you're not seeing other examples to go from, is that `formInput` is designed for cases where you want to use one of the "new" input types, `color`, `date`, etc. Just a regular text input like this, the normal way to do that is to use `formText`.

Of course I do agree it shouldn't be a notice anyway.

csidirop:
Okay I see, thats why I only found `formText`. So I have to change my code. Thanks!

But the issue is still valid.

PR #1048, opened 2025-03-20
Add filter 'all_element_texts'

ebellempire:
This commit adds a new filter for the `all_element_texts` function. I provided some rationale for the addition in this [post on the Omeka Forum](https://forum.omeka.org/t/proposing-a-new-filter-for-all-element-texts/27048). Stylistically, I wasn't sure the best way to write this code (e.g. maybe you all prefer a more compact return statement, maybe it would be better to make some small tweak to `Omeka_View_Helper_AllElementTexts`), so I just did it the simplest way possible. Let me know if you have feedback, etc.


zerocrates:
I don't love the idea of now calling the helper twice for every all_element_texts call. What do you see yourself wanting to change specifically? Things I'm thinking of here...

- Some things you can already do with the existing `display_elements` filter: this is what's used by Hide Elements but it can also do things like rearranging the order of the elements
- Might you instead think about just filtering the name of the partial used to render the output? Changing it from `common/record-metadata.php` to something in your plugin's control would leave you in a partial with the array of texts plus the record, where you could probably accomplish whatever you want without having to look up all the texts twice.

ebellempire:
Below is an example of what I'd like to do (presently relying on the proposed new filter, in place on my dev site), in this case using the unmodified Freedom theme. I have a bunch of custom elements in a custom element set. They are all sort of narrative- or article-related -- things like subtitle, lede, sponsor, etc. Some of them are designed to be displayed in a particular order and/or as a combination (e.g. to create a byline that uses an oxford comma when there are two creators, etc.). 
Screenshot 2025-04-01 at 5 17 01 PM
So in the context of a plugin, how can I override the view at `common/record-metadata.php`? Just adding it to the plugin at `my-plugin/views/public/common/record-metadata.php` (like you might for a theme) doesn't seem to cut it. But in theory this might be the better option.

ebellempire:
It appears that it's not possible to filter the partial name using something like...
```
public function filterAllElementTexts()
{
  $path = _PLUGIN_DIR_.'/views/public/common/record-metadata.php';
  $options = array('return_type'=>'html', 'partial'=>$path);
  $record = 'item';
  return get_view()->allElementTexts($record, $options);
}
```
That always returns a 404 on the item record. Pretty sure the path is valid so maybe there's more to it?

zerocrates:
You can't just provide the same partial in a plugin and have it get used; in Classic the plugins are later in the view stack than the default core public views, so plugins can't override core views like that.

Your later approach is closer to what I'm talking about. Having your plugin provide a view at `views/public/common/record-metadata-override.php` (or any other suitably unique name) and then setting the partial to `common/record-metadata-override.php` in the options array in the call to `all_element_texts` will work today. Then the only component we'd need to add would be a simple filter in the AllElementTexts helper to filter the partial name, i.e. `all_element_texts_partial` so a plugin could trigger usage of its partial without modifying the theme.

A filter on the whole options array just before the options are applied would allow the same thing, plus allow changing any of the other options... but the others aren't particularly interesting, I suppose.

As for your 404, I'd imagine you don't have the filename quite right. The partial path should be the path in your module underneath public/ or shared/.

zerocrates:
I'd suggest your filter look like

```
apply_filters('all_element_texts_options', $options, array('record' => $record));
```

You probably want to pass along the record in case the filter wants to do something that depends on the specific record being displayed.

zerocrates:
So the version as you have it now, have you successfully tried out using it?

ebellempire:
Yeah, it's doing exactly what I needed. Here's a boiled down usage example.
```
filterAllElementTextsOptions($options, $args)
{
  if(
    !is_admin_theme() && 
    isset($args['record']) &&
    isset($args['record']['item_type_id']) &&
    $args['record']->getProperty('item_type_name') == 'Some Type'
  ){ 
      $options['partial'] = 'common/record-metadata-override.php';
  }
  return $options;
}
```
I hadn't even noticed the `common/record-metadata.php` template before; it's a much more convenient place to implement a filter, thanks.

zerocrates:
OK, sounds good. Just wanted to double-check that it works for what you were thinking of doing.

PR #1068, opened 2025-07-30
Create get_display_records().

kimisgold:
This puts display record HTML into an array, while also keeping display_records() for outputting all the HTML.

This can be be checked against [Foundation using the same branch name](https://github.com/omeka/theme-foundation/tree/get-display-records). You can see it in use here: https://github.com/omeka/theme-foundation/blob/35613967773096acfe2eb50357087b4484ce0984/functions.php#L67C1-L90C2.

### DIFF
--- a/application/libraries/globals.php
+++ b/application/libraries/globals.php
@@ -2368,21 +2368,9 @@ function recent_items($count = 10)
     }
     return $html;
 }
-/**
- * Get HTML for records.
- *
- * @since 3.2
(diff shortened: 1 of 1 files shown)


zerocrates:
- I think just do `implode('', $records)`, or even the whole method as just `return implode('', get_display_records...`

- You need the doc comment on get_display_records also

kimisgold:
@zerocrates The PR should be ready to look at with responses to your comments.

PR #1093, opened 2026-02-18
Correct German translations for recent items and collections

ebellempire:
I'm not a German speaker; this change is suggested by a client in Germany and I believe it is correct.


zerocrates:
This came in via the Transifex pull for 3.2.1.

Issue #1103, opened 2026-06-09
Test that landmarks are appropriately labeled across admin


kimisgold:
The [`landmark-labels`](https://github.com/omeka/Omeka/tree/landmark-labels) branch has corrections for various landmarks, mainly `<nav>` elements. Test that they are properly read by screen readers. I'm also taking feedback for vague or inaccurate labels. Here are the affected landmarks:

- Admin side
  - [x] Appearance nav
  - [x] Content nav
  - [x] Admin nav
  - [x] Top search form
  - [x] Items only search form
  - [x] Items nav (across browse, tags, and advanced item search)
  - [x] Header (no longer has extraneous "banner" role)
- Public side
  - [ ] Pagination nav
  - [ ] Footer nav

kimisgold:
I've updated the branch to have labels for both the installation-wide search and the items only search.

aviannamiller:
Feedback for admin-side landmark edits:

- Items search form looks good, although to avoid repetition, I suggest removing the label on the main content area on this page that references the same heading ("Search items"). Up to you if you want to make the label on the search just read "Items", but I understand it's referencing the page and form as a whole, so I do see either argument.
- I see the nav tabs have been wrapped under a nav landmark for Appearance, but in other instances of the same pattern, this is not the case. `#section-nav` tabs appear in settings, user edit, and resource edit pages, to name just those that are part of the core. For the first two, the proper role is there, but no distinguishing label. The latter has a list role, no label.
- For landmarks containing duplicate content, in this case "top pagination" and "bottom pagination", I would recommend making these labels the same (probably just "pagination"?). See [General Principles of Landmark Design, Step 3](https://www.w3.org/WAI/ARIA/apg/practices/landmark-regions/#generalprinciplesoflandmarkdesign). Also, I suppose we could be updating our labels for pagination controls along the same sort of guidance provided here (could be "Search results", "Items browse", etc.). Same for S.

aviannamiller:
For public-side content, the following comments on the pagination and footer nav, followed by additional feedback on other landmarks.

- Pagination labels could be updated to reflect purpose, see previous comment on this.
- Seasons, Santa Fe, Minimalist: labels distinguishing top and bottom pagination should be changed to the same thing for both locations.
[...]

PR #1105, opened 2026-06-16
Populate landmark aria-labels across admin

kimisgold:
Affected landmarks include

* Appearance nav
* Content nav
* Admin nav
* Top search form
* Items only search form
* Items nav (across browse, tags, and advanced item search)
* Header (no longer has extraneous "banner" role)


kimisgold:
Tested in #1103

